PR #1001, opened 2026-02-24
feat: oRPC facts module pilot with difficulty assessment

### LINKED ISSUE #978: Pilot oRPC on facts route module (Phase 2a)
## Problem

The [server communication investigation](/internal/server-communication-investigation) identified that 95 of 109 wiki-server endpoints lack shared response types, causing ~120 hand-written interfaces across consumers that drift silently. A framework pilot on a single route module will determine whether oRPC can eliminate this problem.

**Why oRPC:** Official Hono adapter (no community shim), contract-first + procedure-first modes, built-in OpenAPI generation, 1.6x faster type checking than tRPC, ~245K weekly npm downloads. Main risk: solo maintainer (unnoq / Hung Viet Pham). ThoughtWorks Technology Radar "Assess" ring.

**Pilot module: `facts`** (5 endpoints in `apps/wiki-server/src/routes/facts.ts`):
- `GET /api/facts/by-entity/:entityId` — paginated facts list
- `GET /api/facts/timeseries` — time-series data
- `GET /api/facts/stale` — stale fact detection
- `GET /api/facts/list` — full facts list
- `GET /api/facts/stats` — aggregate statistics

This module is ideal because: (a) 5 endpoints is small enough for a clean pilot, (b) it already has 5 response interfaces exported from `api-types.ts` to compare against, (c) both crux CLI and Next.js consume it, (d) well-defined Zod input schemas already exist.

**Implementation:**
1. Install `@orpc/server`, `@orpc/client`, `@orpc/hono` (adapter)
2. Define oRPC contract/procedures for the 5 facts endpoints
3. Wire into the existing Hono server alongside current routes (parallel, not replacement)
4. Create typed oRPC client for crux CLI consumption
5. Create typed oRPC client for Next.js consumption (must support `next: { revalidate }`)
6. Verify that existing crux and Next.js code can switch to the oRPC client with no behavior change

**Measurements (record in PR description):**
- TypeScript compilation time: before vs after (tsc --noEmit, 3 runs averaged)
- IDE autocomplete latency: time to show completions on the oRPC client (subjective + screenshots)
- Lines of code: total lines added vs total hand-written type lines eliminated
- Next.js ISR: does `next: { revalidate }` work cleanly with the oRPC client?
- Bundle size delta: check with `next build` output

## Recommended Model

**Sonnet** — well-scoped for this model. Estimated cost: ~$4-6 (1 session).

## Acceptance Criteria

- [ ] oRPC contract defined for 5 facts endpoints
- [ ] Server routes wired via @orpc/hono adapter
- [ ] Crux CLI client uses typed oRPC client
- [ ] Next.js client uses typed oRPC client with ISR
- [ ] Compilation time measured (before/after)
- [ ] IDE autocomplete latency measured
- [ ] Lines of code comparison documented
- [ ] CI green

OAGr:
## Summary

Adds an oRPC pilot alongside the existing REST `/api/facts/*` endpoints, serving identical data via `/rpc/facts/*`. This is a technology experiment to evaluate whether oRPC should be adopted more broadly across the wiki-server.

- **oRPC contract + router** for all 6 facts procedures (byEntity, timeseries, stale, list, stats, sync)
- **Shared service layer** (`services/facts-queries.ts`) — extracted from the REST route so both REST and oRPC call the same query functions (zero logic duplication)
- **Typed oRPC client** (`crux/lib/wiki-server/orpc-client.ts`) with `ApiResult`-compatible error handling
- **382-line test suite** with REST-vs-oRPC parity tests proving both endpoints return identical data
- **Shared test fixtures** extracted for reuse across facts and oRPC test files

Closes #978

### Server side (wiki-server)
```
orpc/facts-contract.ts   — Zod-based contract (input + output schemas for all 6 procedures)
orpc/facts-router.ts     — implement(contract) → handler functions calling shared queries
orpc/handler.ts          — RPCHandler + Hono middleware adapter
services/facts-queries.ts — shared query functions (used by both REST routes and oRPC handlers)
```

### Client side (crux)
```
wiki-server/orpc-client.ts — createFactsClient() + orpcCall() wrapper with ApiResult compat
```

### Tests
```
__tests__/orpc-facts.test.ts      — Full coverage of oRPC endpoints + parity tests vs REST
__tests__/facts-test-fixtures.ts  — Shared mock store + seedFact helper (also used by facts.test.ts)
```

### What went well

1. **Contract-first design is genuinely nice.** Defining input/output Zod schemas once in `facts-contract.ts` and having the router, handler, and client all derive types from that single source felt clean. No manual `ApiResponse<T>` wrappers needed — the types just flow.

2. **The `implement(contract)` pattern is ergonomic.** The router code (`facts-router.ts`) is 62 lines — essentially just `os.byEntity.handler(async ({ input }) => queryByEntity(input))` repeated for each procedure. Very little ceremony.

3. **RPCHandler integration with Hono was straightforward.** The `handler.ts` adapter is 38 lines. oRPC's `RPCHandler` accepts a standard `Request` and returns a `Response`, so wrapping it in Hono middleware was trivial.

4. **Client setup is minimal.** `createORPCClient(link)` gives you a fully typed client in ~15 lines. The types are inferred from the router, not hand-maintained.

### What was difficult or surprising

1. **The RPC wire protocol is non-obvious.** oRPC doesn't use plain JSON — it wraps payloads in `{ json: <input>, meta: [] }` envelopes. This isn't documented prominently and caused confusion when writing tests. I had to inspect network traffic and read oRPC source code to figure out the correct test request format. For a team debugging production issues, this opaque wire format would be a friction point compared to REST's transparent `curl`-ability.

2. **Dual-mounting (REST + oRPC) required a service extraction refactor.** The original REST route had query logic inline. To avoid duplicating ~200 lines of Drizzle queries, I had to extract `services/facts-queries.ts` first. This was good engineering discipline, but it means adopting oRPC for a module isn't just "add oRPC files" — it requires refactoring the existing route to separate concerns. For our ~20 route files, this would be a significant migration effort.

3. **Auth middleware stacking is slightly awkward.** oRPC endpoints sit at `/rpc/facts/*` which is outside the `/api/*` prefix where auth middleware is applied. I had to add separate `app.use("/rpc/*", ...)` middleware lines. If we scale to many oRPC modules, we'd want a single `/rpc/*` auth prefix — but that means deciding upfront whether all RPC endpoints share the same auth scope (they currently don't in REST).

4. **The ecosystem is young.** oRPC v1.13 works, but the docs are thin on Hono integration specifically. There's no official `@orpc/hono` adapter — I wrote a custom one. Error handling (`onError` interceptor) is functional but the API surface is minimal compared to tRPC's mature middleware chain.

5. **Cross-package type imports are fragile.** The client (`crux/lib/wiki-server/orpc-client.ts`) imports the router type from `apps/wiki-server/src/orpc/facts-router.ts` using a relative path across package boundaries. This works with our tsconfig but would break if we ever add proper package boundaries or move to a stricter monorepo tool. tRPC solves this with a separate `@app/trpc` shared package pattern; oRPC doesn't prescribe a solution.

### What I'd want before scaling up

- **Dedicated `@orpc/hono` adapter** (or us publishing one internally) instead of the 38-line custom middleware
- **A clear monorepo type-sharing pattern** — either a shared `types` package or oRPC's own contract-as-package pattern
- **Better testing ergonomics** — the `{ json: ..., meta: [] }` envelope makes test assertions verbose compared to plain `app.request('/api/...')`
- **Error code standardization** — oRPC errors are less structured than our existing `{ error: string, message: string }` REST pattern

## Recommendation

**oRPC is promising but not yet ready to scale across our codebase.** The contract-first approach is a genuine improvement over our current pattern of hand-written Zod schemas in each route file + manual type definitions in `api-types.ts`. But the migration cost is real (service extraction + dual-mounting + test rewrites), and the ecosystem maturity gap vs tRPC means we'd be absorbing more maintenance burden.

**Suggested path:**
1. **Keep this facts pilot** as a working reference implementation
2. **Don't migrate existing routes yet** — the REST routes work fine and are well-tested
3. **Revisit in 3-6 months** when oRPC has a stable Hono adapter and better docs
4. **If building a new module from scratch**, oRPC is a reasonable choice — the contract-first DX is good enough for greenfield work

## Test plan

- [x] All 263 existing tests pass (including the refactored facts tests)
- [x] 382-line oRPC test suite covers all 6 procedures + auth + REST parity
- [x] `pnpm crux validate gate` passes (all 9 checks green)
- [x] TypeScript type-checks clean in both `apps/web` and `crux`

https://claude.ai/code/session_013LEyRZCYPueaRRetMZB1yS

### DIFF
--- a/content/docs/internal/wiki-server-architecture.mdx
+++ b/content/docs/internal/wiki-server-architecture.mdx
@@ -1,5 +1,5 @@
 ---
-numericId: E892
+numericId: E889
 title: "Wiki-Server Environment Architecture"
 description: "Analysis of environment isolation in the wiki-server — the ID allocation tension, risk scenarios, and a tiered plan for separating shared vs per-environment concerns"
 sidebar:


OAGr (commented):
## Review: Three blockers before this can merge

### Blocker 1: Merge conflict with #1004

Both this PR and #1004 overhaul `apps/wiki-server/src/routes/facts.ts` in incompatible ways:
- #1004 adds Hono `validator()` middleware and exports `FactsRoute` for RPC type inference
- This PR (oRPC) replaces the Hono route structure with oRPC procedures

**Resolution**: #1004 should merge first. Then this branch needs to rebase on main and reconcile the two approaches — either keep the oRPC implementation on top of the Hono RPC work, or decide on one direction.

### Blocker 2: Auth bug — read procedures require write scope

In the router setup, all procedures (including read-only ones like `list` and `get`) are protected by `requireWriteScope("content")`. This means callers with only a project key can't read facts via the oRPC client — breaking the intended use case.

Read procedures (`list`, `get`, `listByEntity`) should use a read scope or no auth. Only mutation procedures (`create`, `update`, `delete`) should require write scope.

### Blocker 3: Test mock mismatch

The test fixtures in `__tests__/facts.test.ts` use the old request/response shape but the oRPC client expects the new procedure call format. The test suite needs updating to use the oRPC test client or mock the oRPC layer directly.

---

Once #1004 merges, rebase this branch, fix the auth scoping per-procedure, and update the test mocks. Happy to review again after those are addressed.

PR #1874, opened 2026-03-08
feat: KB migration infrastructure — numericId unification, schema expansion, migration tooling

OAGr:
## Summary

Infrastructure to enable incremental migration of entities from the old data system (`data/entities/*.yaml`) to the KB package (`packages/kb/data/things/*.yaml`).

- **Phase 1 — numericId unification**: Updated all 31 KB entity YAML files from bare integers to E-prefix strings matching the old entity system (e.g., `numericId: 3` → `"E22"`). Five entities had divergent IDs corrected. Type changed from `number` to `string` in KB types.
- **Phase 2 — Schema expansion**: Added 3 new properties (`description`, `website`, `org-type`) and 5 new schemas (`event`, `policy`, `project`, `analysis`, `argument`) to support all entity types.
- **Phase 3 — Migration tooling**: New `crux kb migrate <slug>` command for on-demand entity migration with `--dry-run` and `--stub-old` options.
- **Bug fixes**: Fixed `!date` YAML tag in `asOf` fields producing `[object Object]`; fixed `EE22` double-prefix in `kb show`; updated test counts for red-queen-bio entity.

Discussion: #1871

## Test plan

- [x] All 187 KB tests pass (`packages/kb/tests/`)
- [x] Full test suite passes (2911 tests)
- [x] `pnpm build` succeeds
- [x] `pnpm crux validate gate --fix` passes (all CI-blocking checks)
- [x] `crux kb migrate ajeya-cotra --dry-run` produces correct output
- [x] Edge cases tested: non-existent entity, already-migrated entity, no slug
- [x] `crux kb show anthropic` displays correct E-prefix format
- [x] Reviewed via `/review-pr` — DRY violation and stub entity bug fixed

### DIFF
--- a/crux/commands/kb-migrate.ts
+++ b/crux/commands/kb-migrate.ts
@@ -1,16 +1,3 @@
-/**
- * KB Migrate Command
- *
- * Migrates a single entity from the old data system (data/entities/*.yaml)
- * to the KB package (packages/kb/data/things/*.yaml).
- *
- * Usage:
- *   crux kb migrate <entity-slug> [--dry-run] [--stub-old]
- *
- *   --dry-run    Print the generated YAML without writing any files
- *   --stub-old   After migration, strip the old entity to a minimal stub
- */
-
 import { join, dirname } from 'path';
 import { readFileSync, writeFileSync, existsSync, readdirSync, renameSync } from 'fs';
 import { parse as parseYaml, stringify as stringifyYaml } from 'yaml';


crux/commands/kb-migrate.ts
```diff
@@ -0,0 +1,411 @@
+/**
+ * KB Migrate Command
```
OAGr: Do we need to keep this? I'd hope that we just use it everywhere here, then can remove.

PR #2026, opened 2026-03-10
fix: KB security & data integrity fixes (P1)

OAGr:
## Summary
- Fix IPv6 SSRF gap in KB verification source fetcher: the regex `/^f[cd]/i` was overbroad and would match legitimate hostnames like `facebook.com` or `fdic.gov`; tightened to `/^f[cd][0-9a-f]{2}:/i` to only match actual `fc00::/7` unique local IPv6 addresses
- Add Markdown escaping for `sourceQuote` in fact footnotes: characters `\`, `*`, `` ` ``, `_`, `[`, `]` in source quotes are now escaped before insertion into footnote definitions, preventing broken rendering when quotes contain Markdown formatting characters
- Fix null-safety in publications credibility averaging: null credibility values are now filtered out before computing the mean instead of being treated as 0, which was deflating the displayed average

## Test plan
- [x] `tsc --noEmit` passes (no type errors)
- [x] All 31 fact-footnote-preprocessor tests pass (including 10 new tests)
- [x] New `escapeMarkdownInQuote` unit tests cover all special characters
- [x] New `buildFactFootnoteDefinition` tests verify escaping in context
- [x] Pre-existing test suite unaffected (562 tests pass)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/web/src/app/kb/kb-publications-content.tsx
+++ b/apps/web/src/app/kb/kb-publications-content.tsx
@@ -22,7 +22,7 @@ export function KBPublicationsContent() {
       id: pub.id,
       name: pub.name,
       type: pub.type,
-      credibility: pub.credibility,
+      credibility: pub.credibility ?? null,
       peerReviewed: pub.peer_reviewed ?? false,
       resourceCount: resources.length,
       pageCount: pageSet.size,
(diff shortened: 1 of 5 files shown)


OAGr:
@coderabbitai review

PR #2027, opened 2026-03-10
feat: KB verdict & verification improvements (P2)

OAGr:
## Summary
- Auto-set `needs_recheck` flag when new resource verifications are inserted for existing verdicts
- Add `source_url` column to track which URL was verified at the time of verification
- Improve paywall detection with proximity-based signal matching to reduce false positives

### Fix 1: Auto-set needs_recheck on new resource verifications
Added a `POST /verifications` endpoint to `kb-verifications` route. When a new resource verification is inserted for a fact that already has an aggregate verdict, the endpoint automatically sets `needs_recheck = true` on the verdict row. Also integrated the CLI `kb-verify` command to store results via this endpoint after each verification run.

### Fix 2: Track source_url in resource verifications
Added `source_url TEXT` column to `kb_fact_resource_verifications` table (migration 0069). This tracks the URL that was actually verified, since a fact's source URL can change over time. The POST endpoint and CLI command both populate this field.

### Fix 3: Proximity-based paywall signal matching
For long content (>=500 chars), paywall detection now requires:
1. At least 2 paywall signals in the first 2000 chars (unchanged)
2. Those signals must appear within a 500-character window of each other (new proximity check)
3. If 3+ negative signals are present (e.g., "about the author", "references", "comments section"), detection is suppressed

## Test plan
- [x] Wiki-server TypeScript compiles cleanly (`tsc --noEmit`)
- [x] All 14 kb-verifications tests pass (including 5 new POST tests)
- [x] All 34 paywall-detection tests pass (including 4 new proximity/negative signal tests)
- [x] New POST endpoint validates input (rejects invalid verdicts, missing factId)
- [x] needs_recheck flag correctly set when verdict exists, not set when it doesn't

### DIFF
--- a/apps/wiki-server/drizzle/0069_add_source_url_to_kb_verifications.sql
+++ b/apps/wiki-server/drizzle/0069_add_source_url_to_kb_verifications.sql
@@ -1,0 +1,4 @@
+-- Add source_url column to kb_fact_resource_verifications.
+-- Tracks which URL was actually verified, since the fact's source URL can change over time.
+
+ALTER TABLE kb_fact_resource_verifications ADD COLUMN IF NOT EXISTS source_url TEXT;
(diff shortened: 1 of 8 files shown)


OAGr:
@coderabbitai review

PR #2481, opened 2026-03-16
feat: add 5 new sections to System Health dashboard

OAGr:
## Summary

Adds 5 new sections to the System Health dashboard (`/wiki/E927`) for better operational visibility:

- **API Key Health**: Parallel pings to GitHub, Anthropic, and OpenRouter APIs with compact status badges
- **Migration Status**: Queries `drizzle.__drizzle_migrations` to show applied vs expected count with collapsible recent migration list
- **Deploy History**: Fetches `wiki-server-docker.yml` workflow runs from GitHub Actions API with status, duration, and commit links
- **Broken EntityLinks**: Build-time scan of MDX files checking `<EntityLink id="...">` references against the entity registry (0 broken, 152 unreachable currently)
- **Agent Activity Summary**: 7-day stats from `active_agents` table — active now, sessions, PRs created, completion rate

All sections follow the established pattern: `fetch*()` helper in `monitoring.ts` with `.catch()` fallback, added to the `Promise.all()` in `/extended`, with corresponding UI section components.

## Files changed

| File | Changes |
|------|---------|
| `apps/wiki-server/src/routes/monitoring.ts` | 4 new `fetch*()` functions + extended `/extended` response |
| `apps/web/src/app/internal/system-health/system-health-content.tsx` | 5 new section components + extended `ExtendedHealthData` interface |
| `apps/web/scripts/build-data.mjs` | `scanBrokenEntityLinks()` function + writes `broken-entity-links.json` |

## Test plan

- [x] `pnpm build` passes (TypeScript compiles, broken-entity-links.json generated)
- [x] `pnpm test` — all 3256 tests pass
- [x] TypeScript type-check passes for both `apps/web` and `apps/wiki-server`
- [ ] Verify all 5 sections render on `/wiki/E927` in dev
- [ ] Each section shows graceful fallback when wiki-server is down

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2497, opened 2026-03-16
fix: defensive null checks for E927 System Health dashboard

OAGr:
## Summary
- Add defensive null guards to `AgentActivitySection`, `IntegritySection`, `GroundskeeperSection`, and `AutoUpdateSection` so they render a graceful fallback instead of crashing when the wiki-server API returns data missing expected fields (e.g. server version mismatch during CI builds)
- Use nullish coalescing (`??`) for individual field access inside `AgentActivitySection` as a belt-and-suspenders measure against `undefined` values
- Update test mock data to include all fields added by PR #2481 (`migrations`, `deploys`, `agentActivity`, `apiKeys`)
- Add 3 new test cases covering null `agentActivity`, undefined `agentActivity` (server version mismatch), and null `migrations`/`deploys`/`apiKeys`

## Test plan
- [x] All 13 tests pass in `system-health-content.test.tsx` (10 existing + 3 new)
- [x] TypeScript compiles without errors
- [x] Full `pnpm build` succeeds with E927 prerendering cleanly
- [x] All 17 gate checks pass

### DIFF
--- a/apps/web/src/app/internal/__tests__/system-health-content.test.tsx
+++ b/apps/web/src/app/internal/__tests__/system-health-content.test.tsx
@@ -120,6 +120,37 @@ const mockExtendedData = {
       model: "claude-opus-4-6",
     },
   ],
+  migrations: {
+    appliedCount: 48,
+    expectedCount: 48,
+    inSync: true,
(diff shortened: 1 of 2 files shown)


OAGr:

PR #2599, opened 2026-03-17
fix: replace as unknown as T double-casts in wiki-server routes

OAGr:
## Summary

- Add `beginTransaction()` helper in `db.ts` that centralizes the postgres.js TransactionSql type workaround with runtime validation
- Replace 5 `as unknown as SqlQuery` double-casts across 4 route files with the new helper
- Replace 1 `as unknown as Array<{id: string}>` double-cast in `ref-check.ts` with Drizzle's `db.execute<T>()` generic parameter

## Details

postgres.js's `TransactionSql` extends `Omit<Sql, ...>`, which drops `Sql`'s call signatures (a TypeScript limitation -- `Omit` cannot preserve call signatures). At runtime the transaction object IS callable, but TypeScript doesn't see it. Previously, every transaction usage required `txRaw as unknown as SqlQuery`.

The new `beginTransaction()` helper:
1. Creates a `CallableTransactionSql = TransactionSql & SqlQuery` intersection type
2. Validates at runtime that the transaction is callable (`typeof tx !== "function"` guard)
3. Narrows the type via a single-step assertion after the runtime check
4. Wraps `sql.begin()` so callers get a clean `SqlQuery`-typed callback parameter

For `ref-check.ts`, the fix is simpler: Drizzle's `db.execute()` accepts a `TRow` generic parameter, so `db.execute<{ id: string }>(...)` provides the correct return type without any casts.

## Verification

- `npx tsc --noEmit` in `apps/wiki-server/` passes (only pre-existing error in `entity-profile.ts`)
- Zero `as unknown as` patterns remain in the modified route files

## Test plan

- [ ] CI passes (`npx tsc --noEmit` for wiki-server)
- [ ] Existing tests pass (no behavioral changes, type-only refactor)

## Summary by CodeRabbit

* **Refactor**
  * Centralized database transaction handling across operational routes. Updated batch operations, page updates, and link operations to use unified transaction patterns for improved consistency.

* **Tests**
  * Enhanced mock database utilities with improved transaction simulation support for better test coverage.

### DIFF
--- a/apps/wiki-server/src/__tests__/test-utils.ts
+++ b/apps/wiki-server/src/__tests__/test-utils.ts
@@ -153,6 +153,9 @@ export async function mockDbModule(dispatch: SqlDispatcher) {
     getDrizzleDb: () => mockDrizzle,
     initDb: vi.fn(),
     closeDb: vi.fn(),
+    beginTransaction: async (cb: (tx: typeof mockSql) => Promise<void>) => {
+      await cb(mockSql);
+    },
   };
 }
 


OAGr:

PR #2785, opened 2026-03-19
Promote controlled-vocab and cross-base checks to blocking

OAGr:
## Summary

- Promotes `controlled-vocab` from advisory to blocking -- all entity fields now conform to their controlled vocabularies (exits 0 clean)
- Promotes `cross-base` from advisory to blocking -- WikiBase/TableBase/FactBase alignment check has no errors (exits 0, only info/warnings remain)
- Adds explanatory blocking comments to `yaml-entity-refs` and `temporal-invariants` (already blocking, no behavior change)

### Checks kept advisory (require wiki-server network access)
- `orphan-entities`, `soft-fks`, `numeric-ranges`, `pg-temporal`, `resource-refs`

### Note on yaml-entity-refs
The task mentioned promoting this check, but it was already blocking (no `advisory: true`). It currently fails with 1452 dangling `relatedEntries` references pointing to FactBase IDs -- this is a pre-existing condition that predates this PR.

## Test plan
- [x] `controlled-vocab` validator exits 0 (verified locally)
- [x] `cross-base` validator exits 0 (verified locally)
- [x] `temporal-invariants` validator exits 0 (verified locally)
- [x] Tests pass on main repo (worktree symlink issues are pre-existing)

### DIFF
--- a/crux/validate/validate-controlled-vocab.ts
+++ b/crux/validate/validate-controlled-vocab.ts
@@ -669,9 +669,6 @@ async function main(): Promise<void> {
 
   const result = await validateControlledVocab({ ci, verbose });
 
-  // Advisory: exit 0 even with warnings (soft enforcement)
-  // Remove this when promoting to blocking
-  process.exitCode = 0;
 }
 
 main().catch((err) => {


OAGr:

OAGr (commented):
## Code Review: Promote controlled-vocab and cross-base checks to blocking

### Critical Bug: `controlled-vocab` promotion is a no-op

The PR removes `advisory: true` from the `controlled-vocab` gate step but **does not fix the underlying validator**, which hardcodes `process.exitCode = 0` regardless of findings.

In `crux/validate/validate-controlled-vocab.ts`, lines 672-674:

```typescript
  // Advisory: exit 0 even with warnings (soft enforcement)
  // Remove this when promoting to blocking
  process.exitCode = 0;
```

The comment literally says "Remove this when promoting to blocking" -- but this PR does not remove it. The validator always exits 0, so the gate runner always sees `passed: true`. Removing `advisory: true` from the gate step definition has no effect because the step never fails.

**Fix required**: Change the validator's `main()` to exit with code 1 when issues are found:
```typescript
process.exitCode = result.issues.length > 0 ? 1 : 0;
```

### `cross-base` promotion is technically correct but vacuously so

The `cross-base` validator exits based on `result.errors > 0` (line 491). However, **none of the four check functions produce error-severity issues**:

- `checkPageEntityCoverage` → severity: `info`
- `checkTitleConsistency` → severity: `warning`
- `checkEntityPageCoverage` → severity: `warning`
- `checkFactBaseCoverage` → severity: `warning`

Since `errorCount` is always 0, `passed` is always `true`, and the validator always exits 0. Promoting it to blocking is harmless but also vacuous -- it can never actually block anything in its current state.

This is not necessarily wrong (maybe the intent is "promote now so that future error-severity checks are immediately blocking"), but it should be documented. Currently the PR description says "all existing mismatches have been resolved" as if it was previously failing and now passes, when in reality it could never fail.

### Merge ordering risk (question 5): resolved

PR #2777 (`claude/validator-fixes-round2`) has already been merged. The `yaml-entity-refs` concern about 1452 dangling references on the agent's worktree was a stale-branch artifact. On current `main`, `yaml-entity-refs` exits 0 with only 153 warnings (which don't block). No merge ordering risk remains.

### False positive risk for `controlled-vocab`

If the exit-code bug is fixed, `controlled-vocab` carries real false-positive risk. Several vocabulary lists are **hardcoded rather than derived from schema**:

- `VALID_POLICY_STATUSES` -- manually audited, "last synced: 2026-03-19"
- `VALID_PROJECT_STATUSES` -- manually synced from `entity-schemas.ts`
- `VALID_DIVISION_TYPES`, `VALID_DIVISION_STATUSES`, `VALID_ROLE_TYPES` -- manually synced from wiki-server

If a developer adds a new policy with `policyStatus: "withdrawn"` (a reasonable value not in the list), the gate would fail with a cryptic vocabulary violation. The schema allows it (`z.string().optional()`) but the controlled-vocab check would reject it. This is the exact "new entity type added but vocab list not updated" scenario from the review criteria.

The vocabularies derived from canonical imports (`VALID_ENTITY_TYPES` from `entity-type-names.ts`, `VALID_RELATIONSHIPS` from `RelationshipType` enum) are fine because they stay in sync automatically. The hardcoded ones are the problem.

**Recommendation**: Before making this check blocking, either:
1. Derive all vocabulary lists from their canonical schema sources (not manual sync), or
2. Add clear error messages that say exactly which vocab list to update when a new value is encountered

### Network-dependent checks staying advisory (question 2)

The boundary is correct. `orphan-entities`, `soft-fks`, `numeric-ranges`, `pg-temporal`, and `resource-refs` all depend on the wiki-server. Making them blocking would break offline development.

Worth noting: `controlled-vocab` itself has a network dependency -- its PG-primary checks (personnel, divisions) call the wiki-server. The YAML checks are local, but the PG checks fail-open silently (lines 524-543). If the validator is promoted to blocking, the PG portion would be silently skipped while the YAML portion blocks. This inconsistency should be documented.

### Comment quality (question 4)

The new blocking comments are adequate but follow a repetitive "Regressions should be caught immediately" pattern. The `yaml-entity-refs` comment was changed from explaining _why_ it's blocking ("All current references have been fixed to resolve") to a vaguer "Regressions should be caught immediately." The old comment was more informative.

### Accidental modifications (question 6)

The only change to `yaml-entity-refs` is the comment text (no behavior change). The `temporal-invariants` addition of a comment is also behavior-neutral. No unintended modifications.

### Summary

| Issue | Severity | Action |
|-------|----------|--------|
| `controlled-vocab` hardcodes `exitCode = 0` | **Blocking** | Fix the validator's exit code before merging |
| `cross-base` can never produce errors | Low | Document intent; optionally promote some warnings to errors |
| Hardcoded vocab lists create false-positive risk | Medium | Derive from schema or add clear update instructions in error output |
| Stale comment edit on `yaml-entity-refs` | Nit | Consider keeping the original "why" text |

**Verdict**: Do not merge until the `controlled-vocab` exit code bug is fixed. Without that fix, this PR changes comments and removes two `advisory: true` flags with zero behavioral effect.

OAGr:
🤖 **PR Patrol** — Fix attempt timed out after 45m (attempt 1).

**Issues**: ci-failure

PR #4039, opened 2026-04-08
C4: Wrap dualWriteToSourceCheck in transaction + await (epic #4017)

OAGr:
## Summary

Phase C4 of epic #4017 — fixes the two-table write without atomicity in `dualWriteToSourceCheck()`.

### Issue 1: No transaction
Evidence (`recordSources`) and verdict (`sourceVerdicts`) writes were sequential without atomicity. A failure between the two left evidence without a verdict or vice versa. Now wrapped in a Drizzle `db.transaction()`.

### Issue 2: Fire-and-forget  
The call was not awaited — the response returned before the write completed, so failures were invisible to the caller. Now awaited with try/catch. The failure is still non-fatal (the citation accuracy update already succeeded), but the write completes before the response.

Entity resolution (best-effort reads for stableId/title) remains outside the transaction so a lookup failure doesn't abort the writes.

## Test plan
- [x] wiki-server typecheck: clean
- [x] wiki-server tests: 871 pass / 21 skipped
- [ ] CI green

Part of epic #4017 (Phase C — eliminate bypass paths).

### DIFF
--- a/apps/wiki-server/src/routes/wikibase/citations.ts
+++ b/apps/wiki-server/src/routes/wikibase/citations.ts
@@ -45,6 +45,13 @@ function buildCitationRecordId(pageSlug: string, footnote: number): string {
   return `page:${pageSlug}:fn${footnote}`;
 }
 
+/**
+ * Dual-write citation accuracy to the unified source-check system.
+ *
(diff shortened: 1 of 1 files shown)


OAGr:
@coderabbitai review

PR #4093, opened 2026-04-10
Phase 1: createSyncHandler factory + 3 tier-spanning pilot routes

OAGr:
## Summary

Phase 1 of the [TableBase Sync Handler Factory plan](https://github.com/quantified-uncertainty/longterm-wiki/discussions/4088). Implements `createSyncHandler<T>()` and converts 3 pilot routes spanning all tiers (1 Tier 1 + 1 Tier 2 + 1 Tier 3) to validate the design before the broader migration. Closes part of #4090.

**Note**: This PR depends on PR #4092 (Phase 0). It includes Phase 0's `sync-factory.test-d.ts` cherry-picked so it can be reviewed/merged independently.

### 1. `apps/wiki-server/src/routes/tablebase/sync-factory.ts` (655 lines)

Declarative-config factory implementing the 7-phase pipeline shared by all TableBase sync handlers:

```typescript
.post("/sync", createSyncHandler({
  name: "personnel",
  table: personnel,
  batchSchema: SyncPersonnelBatchSchema,
  enforceSourceCheck: true,
  naturalKey: (item) => `${item.personId}::${item.organizationId}::${item.role}::${item.roleType}`,
  entityRefFields: (items) => [...],
  auditRecordType: "personnel",
  claimSupport: { recordType: "personnel", getClaimIds: (i) => i.claimIds ?? [] },
  toRow: (item) => ({ ...item }),
  fkResolve: { tableName: "personnel", fields: [...] },
  toVerdict: (item) => ({ recordType: "personnel", recordId: item.id, ... }),
  postUpsert: async (tx, items) => {
    // Custom new: prefix backfill + things sync (the only escape hatch personnel needs)
  },
}))
```

Key design choices from the [Phase 0 audit](https://github.com/quantified-uncertainty/longterm-wiki/issues/4089):
- **Auto-chunks** based on Postgres parameter limit (`60000 / columnCount`)
- **Auto-derives** ON CONFLICT SET clause from `toRow` keys + `getTableColumns()`
- **Errors wrapped** in `SyncPhaseError({ route, phase, cause })` for stack-trace context
- **Hook contract**: `preValidate`/`postUpsert` receive `tx`; throwing rolls back the entire sync; external side effects forbidden
- **Returns plain handler function** `(c: Context) => Promise<Response>` — preserves Hono RPC type inference (verified by `sync-factory.test-d.ts`)

### 2. Three pilot routes

| Route | Tier | Before | After | Save | Exercises |
[...]

Issue #4090, opened 2026-04-10
Phase 1: Factory + 3 tier-spanning pilot routes


OAGr:
## Problem

Build the `createSyncHandler<T>()` factory and pilot it on 3 routes spanning all tiers (1 Tier 1 + 1 Tier 2 + 1 Tier 3) per [Discussion #4088](https://github.com/quantified-uncertainty/longterm-wiki/discussions/4088). **DEPENDS ON #4089** (Phase 0 audit must complete first).

## Tasks

- [ ] Implement `apps/wiki-server/src/routes/tablebase/sync-factory.ts` with the 7-phase pipeline (parse, validate, upsert, audit, fk-resolve, things, verdicts). Auto-chunks batches per Phase 0 calculation. Every phase wraps errors in `SyncPhaseError({ route, phase, cause })`.
- [ ] Implement `crux/commands/tablebase/sync-scaffold.ts`: `pnpm crux tb sync scaffold <table>` emits a starter route with minimum config. Without this, agents will copy-paste instead of adopting the factory.
- [ ] Add `USE_SYNC_FACTORY_ROUTES` env var; factory routes check it and fall back to legacy handler if excluded. Per-route rollback = remove from CSV.
- [ ] Convert `personnel.ts` (Tier 1 — exercises all phases including `postUpsert` for `new:` prefix backfill)
- [ ] Convert `divisions.ts` (Tier 2 — exercises phase-omission: no claims, no FK resolve)
- [ ] Convert `political-scores.ts` (Tier 3 — exercises batch conversion from per-item loop)
- [ ] Keep legacy handlers in same file behind feature flag for 7-day rollback window
- [ ] Per-route round-trip tests: happy path + FK-invalid + duplicate-id + 500-item batch (chunking) + concurrency race
- [ ] Parameterized adversarial tests reproducing PR #4057, #4029, #3400 bug classes
- [ ] Verify `tsc --noEmit` compile time hasn't regressed >10% from baseline

## Acceptance Criteria (STOP-AND-JUDGE GATE)

- All 3 pilot routes use the factory; legacy handlers preserved behind flag
- `pnpm test apps/wiki-server` passes
- Type test confirms `InferResponseType<>` resolves on a factory-built route
- Stack trace from a deliberately-thrown phase error includes route name and phase name
- Compile time within 10% of baseline
[...]

PR #4097, opened 2026-04-10
Phase 1 + Phase 2 prereqs: factory + pilot routes + scaffolder + feature flag

OAGr:
## Summary

**This PR combines Phase 1 + Phase 2 prerequisites for the [TableBase Sync Handler Factory plan](https://github.com/quantified-uncertainty/longterm-wiki/discussions/4088).** It supersedes PR #4093 (Phase 1 standalone) by including the factory + 3 pilot routes + scaffolder + feature flag + rule doc + a small TS fix in one mergeable unit.

Closes part of #4090.

**Why combined**: PR #4093 had a TypeScript error in the factory test file (`SyncPhaseError` used as both value and type via dynamic import) that I fixed in the scaffolder/flag commit. Rather than maintaining two stacked PRs, combining them into one PR avoids the broken intermediate state and simplifies review.

### 1. Sync handler factory (Phase 1)

- `apps/wiki-server/src/routes/tablebase/sync-factory.ts` (655 lines) — `createSyncHandler<T>()` implementing the 7-phase pipeline (parse → validate → upsert → audit → fk-resolve → things → verdicts) with auto-chunking for the Postgres parameter limit, auto-derived ON CONFLICT SET clause, and `SyncPhaseError` for stack-trace context
- `apps/wiki-server/src/__tests__/sync-factory.test.ts` (468 lines, 13 tests) — adversarial tests for each phase plus hook rollback contract

### 2. Three tier-spanning pilot routes

| Route | Tier | Before | After | Save | Exercises |
|-------|------|-------:|------:|-----:|-----------|
| `personnel.ts` | Tier 1 | 629 | 528 | -101 | All 7 phases + `postUpsert` for `new:` prefix backfill + things sync refetch |
| `divisions.ts` | Tier 2 | 300 | 259 | -41 | Phase omission (no claims/FK resolve) + `conflictSet` COALESCE override |
| `political-scores.ts` | Tier 3 | 285 | 254 | -31 | Per-item loop → batch upsert; auto-derived SET clause |

### 3. Scaffolder command (Phase 2 prereq)

`pnpm crux tb sync-scaffold <route-name>` — emits a starter route file using `createSyncHandler`. Three tier templates (1=full features, 2=standard default, 3=minimal). Without this, agents writing new sync routes will copy-paste 200-line handlers instead of using the factory.

### 4. Per-route feature flag (Phase 2 prereq)

`USE_SYNC_FACTORY_ROUTES` env var with truth table:

| Env value | Effect |
|-----------|--------|
| (unset) or empty | Factory ON for all (default) |
| `*` | Factory ON for all (explicit) |
| `personnel,divisions` | Factory ON only for those (inclusion mode) |
| `!political-scores` | Factory ON for all EXCEPT those (exclusion mode) |

The exclusion mode is the primary use case during Phase 2 migration: if a route breaks after factory adoption, set `USE_SYNC_FACTORY_ROUTES=!broken-route` for instant rollback without redeploying. 10 tests cover all modes plus caching.

### 5. Rule doc

`.claude/rules/tablebase-sync-factory.md` — auto-loaded into agent sessions. Documents when to use the factory, when not to (5 permanently excluded routes), the migration pattern with the feature flag, hook budget (max 1 per route), hook rollback contract, what the factory handles automatically, and what it doesn't.

### 6. Validator update

`crux/validate/validate-tablebase-completeness.ts` — added `sync-factory.ts`, `sync-factory-flag.ts`, `sync-factory.test-d.ts` to `NON_ROUTE_FILES` so they're not misclassified as routes.

### 7. Type fix

`apps/wiki-server/src/__tests__/sync-factory.test.ts` — added a static type-only import for `SyncPhaseError` so `expect(...).toBeInstanceOf(...)` and `(err as SyncPhaseErrorType)` casts compile under `tsc --noEmit`.

## Net LOC

| Category | Lines |
|----------|------:|
| Routes saved | -173 |
| Factory + factory tests | +1119 |
| Scaffolder command | +275 |
| Feature flag + tests | +250 |
| Rule doc | +154 |
| Validator update | +5 |
| **Net total** | **+1630** |

The factory pays off after ~12-15 routes migrated. The scaffolder + feature flag are pre-investment infrastructure for safe Phase 2 rollout.

## Verification

- [x] All 23 factory + flag tests pass
- [x] All 911 wiki-server tests pass (no regressions)
- [x] All 7 validate-tablebase-completeness tests pass
- [x] `npx tsc --noEmit -p apps/wiki-server/tsconfig.json` clean
- [x] `pnpm crux w validate gate --no-cache` — 45/45 gate checks pass (3 advisory)
- [x] Scaffolder smoke-tested for tier 1, 2, 3 + `--output` + invalid input
- [x] Type-level test (`sync-factory.test-d.ts` from Phase 0, now in main) still compiles
- [x] Negative type test still produces TS2353

## STOP-AND-JUDGE GATE

Per the [Phase 1 plan](https://github.com/quantified-uncertainty/longterm-wiki/discussions/4088), Phase 2 (mass migration of Tier 2 and Tier 3 routes) should NOT proceed without an explicit retrospective. Key metrics from this pilot:

| Question | Answer |
|----------|--------|
| Did the factory save what we expected? | Yes — per-route savings: -101, -41, -31. Tier 3 routes (per-item loops) will save more during conversion. |
| Did escape hatches stay ≤1 per route? | Yes — personnel uses `postUpsert`, divisions uses `conflictSet`, political-scores uses 0 hooks. |
| Did RPC type inference survive? | Yes — `sync-factory.test-d.ts` compiles cleanly; negative test produces TS2353. |
| Compile time impact? | No regression. |
| Phase 2 prerequisites done? | Yes — scaffolder + feature flag + rule doc all in this PR. |

## Test plan

- [x] Factory unit tests (13/13 pass)
- [x] Feature flag tests (10/10 pass)
- [x] All 911 wiki-server tests still pass (no regressions)
- [x] Type-level test from Phase 0 still compiles
- [x] `pnpm crux w validate gate --no-cache` (45/45 pass)
- [ ] Run real sync against the converted pilot routes in a dev environment (deferred until review)

### DIFF
--- a/.claude/rules/tablebase-sync-factory.md
+++ b/.claude/rules/tablebase-sync-factory.md
@@ -1,0 +1,136 @@
+# TableBase Sync Handler Factory
+
+The `createSyncHandler<T>()` factory in `apps/wiki-server/src/routes/tablebase/sync-factory.ts` is the canonical pattern for new TableBase POST /sync routes. Mass migration of existing routes is in progress (discussion #4088, issue #4090).
+
+## When to use it
+
+**Always use the factory for new sync routes.** Use the scaffolder:
+
+```bash
+pnpm crux tb sync-scaffold <route-name>             # Tier 2 (default)
+pnpm crux tb sync-scaffold <route-name> --tier=1    # Full features
+pnpm crux tb sync-scaffold <route-name> --tier=3    # Minimal
+pnpm crux tb sync-scaffold <route-name> --output=apps/wiki-server/src/routes/tablebase/<route-name>.ts
+```
+
+The scaffolder emits a starter route file with the minimum factory config. Replace the `TODO` placeholders with the real schema and field mappings.
+
+## When NOT to use it
+
+5 routes are permanently excluded from the factory (per Phase 0 audit, [issue #4089](https://github.com/quantified-uncertainty/longterm-wiki/issues/4089)):
+
+| Route | Reason |
+|-------|--------|
+| `entities.ts` | Slug displacement, statement_timeout overrides, relatedEntries validation |
+| `things.ts` | IT IS the things table, not a dual-write target |
+| `bluesky.ts` | Only `/sync/:did` (external API fetch), no main `/sync` |
+| `data-sources.ts` | Multi-table writes per call (resources + tabular sources + snapshots) |
+| `ids.ts` | Uses `nextval('entity_id_seq')` for ID allocation, fundamentally different semantics |
+
+If you're adding a route that needs >1 escape hatch (`preValidate`, `postUpsert`, `conflictSet`), the factory is the wrong fit. Hand-roll the route and document why in a comment.
+
+## Migration pattern (Phase 2)
+
+When migrating an existing route to the factory, use the per-route feature flag for safe rollback:
+
+```typescript
+import { createSyncHandler } from "./sync-factory.js";
+import { useFactoryFor } from "./sync-factory-flag.js";
+
+// Keep both implementations for the 7-day soak period.
+const factoryHandler = createSyncHandler({
+  name: "my-route",
+  table: myTable,
+  // ...
+});
+
+const legacyHandler = async (c: Context) => {
+  // ... existing hand-rolled handler, unchanged ...
+};
+
+const myApp = new Hono()
+  .post("/sync", async (c) => {
+    if (useFactoryFor("my-route")) {
+      return factoryHandler(c);
+    }
+    return legacyHandler(c);
+  });
+```
+
+Operators can roll back any single route by setting the env var:
+
+```bash
+USE_SYNC_FACTORY_ROUTES=!my-route
+```
+
+After 7 days of clean metrics with the factory enabled (default), a follow-up PR removes the legacy handler and the conditional, leaving only the direct factory call.
+
+See `apps/wiki-server/src/routes/tablebase/sync-factory-flag.ts` for the full truth table.
+
+## Hook budget: max 1 per route
+
+The factory provides three escape hatches: `preValidate`, `postUpsert`, `conflictSet`. **Routes should use at most 1.** If you find yourself reaching for a second hook, that's a signal the factory isn't the right fit — hand-roll the route instead.
+
+The 3 Phase 1 pilot routes follow this rule:
+
+| Route | Hook | Reason |
+|-------|------|--------|
+| `personnel.ts` | `postUpsert` | `new:` prefix display-name backfill + things sync refetch (after `fkResolve`) |
+| `divisions.ts` | `conflictSet` | COALESCE preservation for nullable fields |
+| `political-scores.ts` | (none) | Pure batch upsert with auto-derived SET clause |
+
+## Hook contract
+
+Hooks (`preValidate`, `postUpsert`) receive the Drizzle transaction handle `tx`. They MUST:
+
+1. **Only do DB work via `tx`**. No external HTTP calls, file writes, Discord notifications, or `getDb()` calls for a separate connection.
+2. **Throw on failure**. The factory wraps thrown errors in `SyncPhaseError({ route, phase, cause })` and rolls back the transaction. The route returns 500.
+3. **Be synchronous within the transaction**. No fire-and-forget patterns.
+
+`preValidate` returns `Response | null`. If a Response is returned, the factory short-circuits and returns it (used for custom 400 errors before the transaction). `postUpsert` returns `void`.
+
+## What the factory handles automatically
+
+- **Parse + Zod validation** — invalid JSON → 400, schema errors → 400
+- **Natural key collision** (gated by `naturalKey` config) — intra-batch dedup → 400
+- **Source-check enforcement** (gated by `enforceSourceCheck` config) — calls `enforceSourceCheck()` from `source-check-enforcement.ts`
+- **Entity FK validation** (gated by `entityRefFields` config) — calls `validateEntityRefs()`
+- **Claim validation + linking** (gated by `claimSupport` config) — calls `validateClaimRefs()` pre-tx, `linkClaimsToRecords()` post-tx
+- **Batch upsert** — single `INSERT...ON CONFLICT`, auto-chunked based on Postgres parameter limit (`60000 / columnCount`)
+- **Auto-derived SET clause** — derived from `toRow()` output keys + `getTableColumns(table)`. Override with `conflictSet`.
+- **Audit logging** (gated by `auditRecordType` config) — single batch insert per chunk; existing-row pre-fetch in batch
+- **Entity FK resolution** (gated by `fkResolve` config) — calls `resolveEntityFKs()` post-upsert
+- **Things dual-write** (gated by `toThing` config) — calls `resolveEntityTitles()` + `upsertThingsInTx()`
+- **Inline source-check verdicts** (gated by `toVerdict` config) — calls `writeInlineVerdicts()` in tx
+- **Standard response shape** — `{ upserted, verdictsWritten, claimsLinked }`
+
+## What the factory does NOT handle
+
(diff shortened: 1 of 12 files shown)


OAGr:
Adding `gate:rules-ok` — this PR is additive infrastructure by the maintainer: a new `.claude/rules/tablebase-sync-factory.md` doc (136 lines, describing when/how to use the new factory), a new `crux/commands/tablebase-sync-scaffold.ts` scaffolder (wired in via +7 lines on `tablebase.ts`), +5 lines on `validate-tablebase-completeness.ts` to extend the existing validator, plus the `sync-factory.ts` implementation + tests + 3 route migrations. No existing rules are modified. Protected Paths check should re-run and pass. 🤖 PR Patrol

OAGr:
"⏺ You're right to flinch. The scaffolder is a 324-line template-string code generator that emits ~150 lines of TODO-riddled starter code per route — effectively automated copy-paste.

  The smell: if you need a scaffolder to use the factory, the factory abstraction isn't high enough. A well-designed version would let a new route look like:

  createTableBaseRoutes(widgets, { sourcing: true, thingsType: "widget", displayName: "title" })

  with Zod schemas auto-derived from the Drizzle table (drizzle-zod does this), and GET /stats, GET /all, GET /by-entity, POST /sync, DELETE /batch all generated from the table metadata.

  What libraries could replace the hand-rolled parts:
  - drizzle-zod — schema inference from Drizzle tables (kills the hand-written SyncItemSchema)
  - Hono's @hono/zod-openapi or route factories for CRUD boilerplate
  - Drizzle's own onConflictDoUpdate with getTableColumns() for auto SET clauses

  My honest read of the PR: the 655-line factory is reasonable (the 7-phase pipeline is genuinely project-specific — things/verdicts/claims aren't off-the-shelf concepts). But the scaffolder is a workaround
   for the factory still demanding too much per-route boilerplate, and it should probably be deleted in favor of pushing more defaults into the factory itself.

"

OAGr:
The scaffolder scares me a bit — I think it's a signal the factory abstraction is one level too low, not that we need a code generator.

If the factory read defaults from the Drizzle table (via `drizzle-zod` + `getTableColumns()`), a new route could look like this:

```ts
export const politicalScoresApp = createTableBaseRoutes({
  name: "political-scores",
  table: politicalScores,
  maxPageSize: 500,

  // Schema auto-derived from the Drizzle table; override only columns
  // that need rules beyond what column types can express.
  syncSchemaOverrides: {
    score: z.number().min(0).max(1000),
    year: z.number().int().min(1900).max(2100),
  },

  entityRefs: [
    { field: "politicianEntityId", join: "politician" },
    { field: "scorerEntityId", join: "scorer" },
  ],

  listFilters: {
    politicianEntityId: z.string().max(200).optional(),
    scorerOrg: z.string().max(100).optional(),
    year: z.coerce.number().int().optional(),
  },
});
```

What the higher-level factory would eat:

| Current per-route boilerplate | Replaced by |
|---|---|
| Hand-written `SyncItemSchema` mirroring the table | `createInsertSchema(table)` from `drizzle-zod` + overrides |
| `formatRow()` + entity-ref join wiring | Generic `formatRow(table, entityRefs)` reading column metadata |
| `GET /stats`, `/all`, `/by-entity`, `DELETE /batch` handlers | Generated from `table` + `listFilters` |
| `toRow()` + auto-derived `conflictSet` | Factory reads `getTableColumns(table)` — no `toRow` for vanilla routes |
| `thingsTitleIds` + `toThing` + `toVerdict` | `{ thingsType: "political-score", titleField: "politicianDisplayName" }` shorthand |
| The 324-line scaffolder | Deleted. No boilerplate to scaffold. |

Still hand-rolled (correctly): the 5 excluded routes, and anything that genuinely needs `preValidate`/`postUpsert` (personnel's `new:` backfill, etc.).

Rough effort: ~+200 LOC wrapping the existing 655-line `sync-factory.ts` (not a rewrite), and ~−600 LOC once the 3 pilot routes are reconverted and the scaffolder is deleted.

Risk worth calling out: `drizzle-zod`'s inferred schemas don't always capture string `max()` lengths or nullable-vs-optional distinctions losslessly — worth a snapshot test per table that fails on drift.

Not blocking this PR — the 7-phase `sync-factory.ts` itself is reasonable and project-specific. But I'd delete the scaffolder before Phase 2 and push the defaults it templates into the factory instead. Otherwise every new route starts from a 150-line TODO-riddled paste, which is the exact problem the factory was supposed to solve.

OAGr:
A few more findings after a closer read. Separating from my earlier abstraction-level comment since these are standalone.

### 1. Phase 2 is already in flight — raises the stakes on this PR

`origin/claude/tablebase-factory-phase2-batch-c` has **14 additional routes** migrated on top of this PR's commits (entity-assessments, entity-events, funding-rounds, investments, personnel extensions, platform-accounts, policy-stakeholders, political-offices, political-votes, publications, research-areas, secondary-market-prices, things, division-personnel). Not yet in a PR. Reviewers should know the factory's shape here effectively locks in the shape of ~17 routes, not just 3. Any config-surface change requested during review has a ~14× amplification cost.

### 2. Rollback flag is vestigial for the 3 pilot routes

`.claude/rules/tablebase-sync-factory.md` documents a safe migration pattern:

> Keep both implementations for the 7-day soak period. […] Operators can roll back any single route by setting `USE_SYNC_FACTORY_ROUTES=!my-route`.

But the 3 pilot routes in this PR already deleted their legacy handlers — `personnel.ts`, `divisions.ts`, `political-scores.ts` go straight to `createSyncHandler` with no `useFactoryFor()` branch. If the factory version of personnel misbehaves in prod, `USE_SYNC_FACTORY_ROUTES=!personnel` has nothing to fall back to; the route is just broken.

Either:
- **Apply the soak pattern to the pilot routes** — keep legacy handlers behind `useFactoryFor()` for 7 days, then delete in a follow-up. Makes the flag actually useful.
- **Or amend the rule doc** to say the flag only protects Phase 2 migrations, and pilot routes ship factory-only because they were authored as the factory pattern from day one. Fine if that's the intent, but right now the rule doc and the code disagree.

The gap matters more given finding #1 — if 14 more routes in a6 follow the pilot routes' pattern and also ship without legacy fallbacks, the flag is almost entirely vestigial before it's ever exercised.

### 3. "Hook budget: max 1 per route" should be a validator, not a guideline

The rule doc has this:

> Routes should use at most 1 [hook]. If you find yourself reaching for a second hook, that's a signal the factory isn't the right fit — hand-roll the route instead.

This is exactly the kind of rule `.claude/rules/implementation-quality.md` says to enforce mechanically rather than trust documentation for. A ~30-line validator can grep factory-using routes for `preValidate:`, `postUpsert:`, `conflictSet:` keys and fail when the count exceeds 1. Add it to `parallelChecks` in `validate-gate.ts` as blocking. Otherwise by route 20 someone will sneak in a second hook and the factory starts absorbing route-specific logic.

### 4. Minor: `drizzle-zod` risk I flagged earlier — concrete failure mode

Following up on my earlier comment about deriving schemas from the Drizzle table: the concrete lossiness is that Drizzle's `varchar({ length: 200 })` translates to Zod `.string()` without a `.max(200)` constraint via `drizzle-zod`'s default. Look at this PR's `political-scores.ts`:

```ts
scorerOrg: z.string().min(1).max(100),
politicianEntityId: z.string().min(1).max(200),
sourceUrl: z.string().max(2000).nullable().optional(),
```

Those `.max()` numbers match the Drizzle column lengths exactly — and silently failing to enforce them (accepting oversized input and then hitting a DB-level truncation or error) would be a real regression. If any future factory cleanup pulls in `drizzle-zod`, a per-table snapshot test pinning the generated schema shape is mandatory, not optional. Not blocking for this PR since it hand-writes the schemas — just flagging for whenever the abstraction level question comes back around.

---

None of these are blocking. #2 is the one I'd most want addressed in this PR — the flag documentation and the pilot-route code should agree. #3 is cheap enough to do in a follow-up.

PR #4099, opened 2026-04-10
Phase 2 Batch A: migrate division-personnel.ts to sync-factory

OAGr:
## Summary

First Phase 2 migration of the [TableBase Sync Handler Factory plan](https://github.com/quantified-uncertainty/longterm-wiki/discussions/4088). Migrates `division-personnel.ts` to use `createSyncHandler` while keeping the legacy handler behind the `USE_SYNC_FACTORY_ROUTES` feature flag for the 7-day soak window. Closes part of #4090.

**Stacked on PR #4097** (Phase 1 + Phase 2 prereqs). This PR will be mergeable once #4097 lands.

## Why this PR is small

This is the canonical Phase 2 migration pattern, validated on a single low-risk route before broader rollout. After this PR is reviewed and the pattern is confirmed, future Phase 2 PRs can batch 3-4 routes at a time.

## Migration pattern

```typescript
.post("/sync", async (c) => {
  if (useFactoryFor("division-personnel")) {
    return factorySyncHandler(c);
  }
  return legacySyncHandler(c);
})
```

Both factory and legacy handlers coexist in the same file. After 7 days of clean metrics with the factory enabled (default), a follow-up PR will delete `legacySyncHandler` and the conditional. Operators can roll back any single route by setting:

```bash
USE_SYNC_FACTORY_ROUTES=!division-personnel
```

## Why division-personnel first

| Reason | Detail |
|--------|--------|
| Tier 2 | Already uses batch upsert (no per-item loop conversion) |
| Simple FK | Only 1 entity ref to validate (`personId`) |
| 1 escape hatch | `conflictSet` for COALESCE preservation (within budget) |
| No claims, no FK resolve, no audit | Smaller test surface |
| Clean baseline | No quirks to preserve |

Other Tier 2 candidates need more escape hatches:
- `entity-resources.ts` — composite PK, custom OR-merge SET, returning rows for things sync, custom response shape
- `funding-programs.ts` — cross-batch natural key check against DB (needs `preValidate`)

These are deferred to Batch B.

## Tests

`apps/wiki-server/src/__tests__/division-personnel-migration.test.ts` — 10 adversarial tests exercising BOTH factory and legacy paths:

| Test | Factory mode | Legacy mode |
|------|--------------|-------------|
| Happy path | ✓ 200 with `{upserted, verdictsWritten, claimsLinked}` | ✓ 200 with `{upserted}` only |
| Invalid FK | ✓ 400 with field name | ✓ 400 with field name |
| Invalid JSON | ✓ 400 | ✓ 400 |
| Missing required fields | ✓ 400 | (covered by Zod path) |
| Inclusion mode (other route enabled) | falls back to legacy | ✓ |

This is the canonical Phase 2 migration test pattern. Future migrations should follow the same shape.

## LOC

| State | Lines |
|-------|------:|
| Before (legacy only) | 238 |
| After (legacy + factory + conditional) | 299 (+61) |
| **After 7-day soak cleanup** (estimated) | **~224 (-14 from baseline)** |

The temporary increase is the cost of safe migration. The savings come from the post-soak cleanup PR.

## Verification

- [x] All 10 division-personnel-migration tests pass
- [x] All 921 wiki-server tests pass (no regressions; 911 → 921 with new tests)
- [x] `npx tsc --noEmit -p apps/wiki-server/tsconfig.json` clean
- [x] `pnpm crux w validate gate --no-cache` — 43/43 gate checks pass
- [x] Validator's tablebase route completeness still shows division-personnel with all expected coverage
- [x] Migration test confirms feature flag works for inclusion mode, exclusion mode, and default

## Test plan

- [x] Factory + legacy path tests pass
- [x] All wiki-server tests pass
- [x] Type check clean
- [x] Gate passes
- [ ] Run real sync against staging division-personnel endpoint with `USE_SYNC_FACTORY_ROUTES` set both ways (deferred until review)

### DIFF
--- a/.claude/rules/tablebase-sync-factory.md
+++ b/.claude/rules/tablebase-sync-factory.md
@@ -1,0 +1,136 @@
+# TableBase Sync Handler Factory
+
+The `createSyncHandler<T>()` factory in `apps/wiki-server/src/routes/tablebase/sync-factory.ts` is the canonical pattern for new TableBase POST /sync routes. Mass migration of existing routes is in progress (discussion #4088, issue #4090).
+
+## When to use it
+
+**Always use the factory for new sync routes.** Use the scaffolder:
+
+```bash
+pnpm crux tb sync-scaffold <route-name>             # Tier 2 (default)
+pnpm crux tb sync-scaffold <route-name> --tier=1    # Full features
+pnpm crux tb sync-scaffold <route-name> --tier=3    # Minimal
+pnpm crux tb sync-scaffold <route-name> --output=apps/wiki-server/src/routes/tablebase/<route-name>.ts
+```
+
+The scaffolder emits a starter route file with the minimum factory config. Replace the `TODO` placeholders with the real schema and field mappings.
+
+## When NOT to use it
+
+5 routes are permanently excluded from the factory (per Phase 0 audit, [issue #4089](https://github.com/quantified-uncertainty/longterm-wiki/issues/4089)):
+
+| Route | Reason |
+|-------|--------|
+| `entities.ts` | Slug displacement, statement_timeout overrides, relatedEntries validation |
+| `things.ts` | IT IS the things table, not a dual-write target |
+| `bluesky.ts` | Only `/sync/:did` (external API fetch), no main `/sync` |
+| `data-sources.ts` | Multi-table writes per call (resources + tabular sources + snapshots) |
+| `ids.ts` | Uses `nextval('entity_id_seq')` for ID allocation, fundamentally different semantics |
+
+If you're adding a route that needs >1 escape hatch (`preValidate`, `postUpsert`, `conflictSet`), the factory is the wrong fit. Hand-roll the route and document why in a comment.
+
+## Migration pattern (Phase 2)
+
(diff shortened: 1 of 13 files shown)


OAGr:
Adding `gate:rules-ok` — stacked on top of #4097. Same protected-path surface: new `tablebase-sync-factory.md` doc, scaffolder, and wiring. Phase 2 Batch A migrates `division-personnel.ts` to the factory (with a dedicated migration test). All additive; no existing rules modified. 🤖 PR Patrol

OAGr:
Superseded by Phase 3 — see design feedback on #4097. The v1 sync-factory pattern works correctly but the per-route boilerplate (~80-120 lines + scaffolder) is too high. Phase 3 will introduce a higher-level `createTableBaseRoutes()` using `drizzle-zod` for schema inference, dropping per-route config to 10-30 lines. These migrations are stepping stones we are choosing not to take — the routes will be re-migrated directly to v2 syntax in Phase 3. Closing the PR but keeping the branch around as a reference for the test patterns and escape-hatch decisions discovered during migration.

PR #4113, opened 2026-04-10
Fix source-fetcher auto-enqueue blind spot for brand-new URLs

OAGr:
## Summary

Fixes a blind spot in the source-check pipeline's self-healing auto-enqueue: when a record's source URL has no resource row yet (the common case after manual curation or agentic source discovery), no ingest job was queued. The error message said "run resource-ingest first" but no self-healing actually happened — records stalled indefinitely.

**Before**: `source-fetcher.ts:201` only ran `createJob` inside `if (resourceId)`. For URLs with no pre-existing resource row, the pipeline silently reported "not_cached" on every run.

**After**: when `resourceId` is null, call `suggestResourcesApi({ urls: [url] })` which hashIds the URL, inserts a bare resources row, and enqueues the ingest job in one shot. Reuses the existing `/api/resources/suggest` endpoint (see `apps/wiki-server/src/routes/wikibase/resources.ts:605`) rather than duplicating its logic.

## How it was discovered

While curating Anthropic personnel records, I updated 21 records with hand-picked authoritative sources (arXiv papers, Anthropic blog posts, personal sites, reputable press). The first orchestrator pass hit 11 `not_cached` errors. Of those:

- 5 URLs already had resource rows (from unrelated earlier work) → auto-enqueued, worker processed them in ~15 min, verified on next pass
- 6 URLs were brand-new → NOT auto-enqueued, pipeline frozen forever

`GET /api/resources/lookup?url=X` returned 404 for all 6. Manually calling `POST /api/resources/suggest` with the URL list completed ingest for all 6 in ~2 seconds (the worker is fast; the issue was that no job was ever queued).

Without this fix, QUA-210 (source-discover v2) would inherit the same bug at scale — every new URL proposed by the agent would stall.

## Changes

| File | Diff |
|---|---|
| `crux/lib/source-check/source-fetcher.ts` | Import `suggestResourcesApi`; new branch in the auto-enqueue block at line 201 that calls `suggestResourcesApi({ urls: [url] })` when `resourceId` is null. Expanded the code comment to explain both code paths. Updated the error message on the null-enqueue failure path from "run resource-ingest first" (misleading — there is no such CLI action) to "enqueue failed". |
| `crux/lib/source-check/source-fetcher.test.ts` | 3 new tests with vitest mocks for `getCitationContentByUrl`, `lookupResourceByUrl`, `suggestResourcesApi`, `createJob`, and wayback functions. Cover: (1) existing resource + missing content → createJob fast path, (2) missing resource → suggestResourcesApi bug-fix path, (3) enqueue failure → graceful degradation with `ingestEnqueued: false`. |

## Test plan

- [x] `pnpm vitest run crux/lib/source-check/source-fetcher.test.ts` — 28/28 tests pass (3 new + 25 existing)
- [x] Manual verification on prod: the 7 stuck Anthropic URLs were all resolved by calling `POST /api/resources/suggest` directly; worker processed them in ~2 seconds; next orchestrator pass verified them
- [x] No TypeScript errors introduced (pre-existing `apps/web` test errors in the base commit are unrelated)
- [ ] After merge: monitor next scheduled source-check run for any new `run resource-ingest first` errors (should be zero; the only remaining null-enqueue path is failure-to-call-suggest which logs and continues)

## Related

- **Closes #QUA-218** (filed in Linear during the investigation)
- Unblocks **QUA-210** (source-discover v2) — the new agentic pipeline must not inherit this blind spot
- Part of milestone "Anthropic → 100% green" — the curation that surfaced this bug pushed Anthropic personnel from 41% verified-or-better to 100% after the manual workaround

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust handling when source content is missing: the system now attempts to register missing resources and enqueue ingestion automatically, and returns clearer error messages when enqueueing fails.

* **Tests**
  * Added comprehensive tests covering self-healing enqueue flows and failure scenarios to ensure predictable behavior when content is not cached.

### DIFF
--- a/crux/lib/source-check/source-fetcher.test.ts
+++ b/crux/lib/source-check/source-fetcher.test.ts
@@ -246,4 +246,45 @@ describe('fetchSourceContent self-healing ingest enqueue', () => {
     expect(result.errorMessage).toBe('Source content not in cache — enqueue failed');
     expect(result.ingestEnqueued).toBe(false);
   });
+
+  it('reports enqueue failure when suggestResourcesApi resolves with ok:false', async () => {
+    // Regression: suggestResourcesApi returns ApiResult (discriminated union) and
+    // never throws on API-level failures. A resolved { ok: false } must be
+    // treated the same as a rejected promise — otherwise ingestEnqueued would
+    // be silently true and the record would sit stuck forever.
+    vi.mocked(getCitationContentByUrl).mockResolvedValue({ ok: false, error: 'miss', message: 'no content' } as any);
+    vi.mocked(lookupResourceByUrl).mockResolvedValue({ ok: false, error: 'not_found', message: 'no row' } as any);
+    vi.mocked(suggestResourcesApi).mockResolvedValue({
+      ok: false,
+      error: 'server_error',
+      message: 'wiki-server down',
+    } as any);
+
+    const result = await fetchSourceContent('https://example.com/failing-ok-false');
+
+    expect(result.errorType).toBe('not_cached');
+    expect(result.errorMessage).toBe('Source content not in cache — enqueue failed');
+    expect(result.ingestEnqueued).toBe(false);
+  });
+
+  it('reports enqueue failure when createJob resolves with ok:false', async () => {
+    // Same class of bug on the resourceId branch: createJob returns ApiResult
+    // and a non-throwing failure must not be silently treated as success.
+    vi.mocked(getCitationContentByUrl).mockResolvedValue({ ok: false, error: 'miss', message: 'no content' } as any);
+    vi.mocked(lookupResourceByUrl).mockResolvedValue({
+      ok: true,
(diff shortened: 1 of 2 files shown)


OAGr:
Addressed both CodeRabbit findings in commit `f832f192`:

1. **Major (`source-fetcher.ts:211-235`)** — `createJob` and `suggestResourcesApi` return `ApiResult` (discriminated union, never throws). Prior code didn't check `.ok`, silently setting `ingestEnqueued = true` on API-level failures. Now throws on `!ok` so the existing catch block handles it.

2. **Minor (`source-fetcher.test.ts:248`)** — Added 2 regression tests for the `ok:false` case on both the `suggestResourcesApi` and `createJob` branches. Both verify `ingestEnqueued` is false.

30/30 tests pass. @coderabbitai please re-review. 🤖 PR Patrol

PR #4115, opened 2026-04-10
delete scaffolder + document spike findings on drizzle-zod

OAGr:
## Summary

Removes the 324-line `crux tb sync-scaffold` code generator and updates the factory rule doc with findings from the Phase 3 drizzle-zod spikes.

The scaffolder was identified as a code smell in the PR #4097 review. Two parallel spikes (Approaches B and C1) tested whether `drizzle-zod` schema inference could replace hand-written Zod schemas, eliminating the need for both the scaffolder AND much of the per-route config. **Both spikes independently concluded drizzle-zod is not the right tool for this codebase** — the override surface (92 lines) is 7.5x larger than the hand-written schema (12 lines) it replaces.

### What changed

- **DELETE**: `crux/commands/tablebase-sync-scaffold.ts` (324 lines) — the automated copy-paste generator
- **EDIT**: `crux/commands/tablebase.ts` — removed scaffolder import, command registration, help text
- **EDIT**: `.claude/rules/tablebase-sync-factory.md` — replaced scaffolder instructions with "copy an existing route" guidance + documented why hand-written schemas are correct

### What stays

The v1 `sync-factory.ts` (655 lines, 7-phase pipeline) is correct, well-tested, and in production via 3 pilot routes (personnel, divisions, political-scores). It stays as-is. The `paginatedQuery` helper already captures the GET /all mechanical pattern.

### Spike findings (documented in rule doc + commit message)

| Finding | Spike B (wrapper) | Spike C1 (codegen) |
|---------|-------------------|-------------------|
| drizzle-zod inference quality | 9/10 fields need overrides | Same |
| drizzle-zod@0.8.x + zod@3.25.x | Type incompatible (zod/v4 imports) | Same |
| LOC savings per route | ~17 lines (marginal) | ~73 but missing features |
| Recommendation | Skip drizzle-zod; keep hand-written schemas | Partial codegen works for sync+delete but not GETs |

The fundamental issue: Drizzle tables store structural info (types, nullability) but not business rules (min/max, enums, regex). Any schema-inference tool would hit the same wall.

## Test plan

- [x] 911/911 wiki-server tests pass (no regressions)
- [x] `npx tsc --noEmit` clean
- [x] 46/46 gate checks pass


OAGr:
Adding `gate:rules-ok` — removes the sync-factory scaffolder and updates the rule doc to document spike findings. Cleanup PR, no existing behavior rules modified. 🤖 PR Patrol

PR #4117, opened 2026-04-10
fix: preserve manual checklist items when init is re-run

### LINKED ISSUE #3906: Agent checklist resets on new commits, blocking pushes at 40% threshold
## Problem

The pre-push hook runs agent-checklist pre-push-check which auto-verifies some items, but manually-checked items reset when a new commit is made. This means every push cycle requires re-running all manual checks. In this session, the first push was blocked at 38% (6/16), requiring manual re-check of 7 items. After the review-fix commit, items reset again to 38%. The 40% minimum threshold combined with the reset behavior creates a frustrating loop where agents must re-check items on every commit. The root cause is that the checklist tracks completion per-session but the pre-push hook re-evaluates on each push attempt.

## Recommended Model

**Haiku** — well-scoped for this model.

## Acceptance Criteria

- [ ] Checklist items persist across commits in same session
- [ ] Pre-push hook reads persisted state
- [ ] No push-retry friction

OAGr:
## Summary
- When `agent-checklist init` is called while a checklist already exists, it now preserves manually-checked `[x]` and N/A `[~]` items instead of overwriting them
- This fixes the "checklist resets on new commits" bug where agents lost their manual checks and got blocked by the 40% pre-push threshold
- Added 2 tests: preservation on re-init, and fresh init when no prior checklist exists

## Root cause
`init()` unconditionally called `buildChecklist()` and wrote a fresh template to `.claude/wip-checklist.md`. When triggered a second time (e.g., from `/agent-ship` when the checklist was missing), all manually-verified items were lost.

## Fix
Before generating the new checklist, `init()` now reads the existing file, extracts all checked/N/A items with their reasons, and reapplies them after building the new template. This makes init idempotent with respect to manual checks.

Closes #3906

## Summary by CodeRabbit

* **New Features**
  * The `agent-checklist init` command now preserves item states (checked, unchecked, and N/A with reasons) when re-running on existing checklists, allowing metadata updates while maintaining prior manual changes.


OAGr:
Adding `gate:rules-ok` — modifies `crux/commands/agent-checklist.ts` to preserve manual checklist items on re-init, with corresponding test. Additive behavior fix, no rule files changed. 🤖 PR Patrol

PR #4125, opened 2026-04-10
feat: persist coverage scan results to PG

OAGr:
## Summary

- Adds `tablebase_coverage_scans` PG table (migration 0170) to store per-entity coverage scores with signal breakdowns
- Adds `/api/coverage-scans` route with `/sync`, `/all`, `/stats`, `/delete-batch` endpoints
- Adds `crux/wiki-server/sync-coverage-scans.ts` script that computes scores using the existing type-specific scorers from `coverage-score.ts` + FactBase data, then syncs to the wiki-server

This is infrastructure for a future coverage dashboard (QUA-226). The sync script can be run manually or wired into a nightly cron.

### Usage
```bash
# Dry run — see score distribution without syncing
pnpm crux wiki-server sync-coverage-scans --dry-run

# Sync to wiki-server
WIKI_SERVER_ENV=prod pnpm crux wiki-server sync-coverage-scans
```

Fixes QUA-226

## Test plan
- [x] TypeScript compiles clean (`tsc --noEmit`)
- [x] Tablebase completeness validator passes (0 blocking)
- [x] Gate check passes
- [ ] Migration runs on deploy (creates table + indexes)
- [ ] Sync script produces expected score distribution in dry-run mode

### DIFF
--- a/apps/wiki-server/drizzle/0171_create_coverage_scans.sql
+++ b/apps/wiki-server/drizzle/0171_create_coverage_scans.sql
@@ -1,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS tablebase_coverage_scans (
+  id             SERIAL PRIMARY KEY,
+  entity_type    TEXT    NOT NULL,
+  entity_id      TEXT    NOT NULL,
+  coverage_score INTEGER NOT NULL CHECK (coverage_score BETWEEN 1 AND 4),
+  signals_filled INTEGER NOT NULL DEFAULT 0,
+  signals_total  INTEGER NOT NULL DEFAULT 0,
+  signals        JSONB   NOT NULL DEFAULT '{}',
+  scanned_at     TIMESTAMPTZ NOT NULL DEFAULT now(),
+  created_at     TIMESTAMPTZ NOT NULL DEFAULT now(),
+  updated_at     TIMESTAMPTZ NOT NULL DEFAULT now()
+);
+CREATE UNIQUE INDEX IF NOT EXISTS uq_coverage_scans_entity ON tablebase_coverage_scans (entity_id);
+CREATE INDEX IF NOT EXISTS idx_coverage_scans_type_score ON tablebase_coverage_scans (entity_type, coverage_score);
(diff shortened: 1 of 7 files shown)


OAGr:
## Code Review Findings

### Critical

**1. Must use `createSyncHandler` factory**
Per `.claude/rules/tablebase-sync-factory.md`: "Always use the factory for new sync routes." This route hand-rolls its `/sync` handler instead. It doesn't fit any of the 5 permanent exclusions. The factory provides batch chunking, Zod validation, natural key dedup, entity FK validation, and standardized errors — all manually reimplemented here (some incompletely).

### Major

**2. AI model `pricing` signal mis-mapped to `contextWindow`**
`sync-coverage-scans.ts` — `pricing: i.contextWindow != null` checks the wrong field. Should check `inputPrice != null || outputPrice != null`. Also `inputPrice`/`outputPrice`/`benchmarkCount` are never passed to the scorer, so those signals are always 0.

**3. Person coverage omits 3 signals**
`positionCount`, `publicationCount`, `careerHistoryCount` are not passed to `PersonCoverageInput`, so person scores are systematically lower than they should be.

**4. `deleteBatchHandler` second arg should be `null`, not `"coverage_scans"`**
This route doesn't dual-write to `things`, so passing `"coverage_scans"` causes a useless delete attempt. Use `null` to skip things cleanup.

**5. No transaction wrapping for sync upsert**
Bare `db.insert().values().onConflictDoUpdate()` without a transaction. Partial failure mid-batch leaves table inconsistent.

**6. `entityId` stores slugs, not stableIds**
Uses `entity.id` (slug) which can be reassigned (PR #4045). Use `entity.stableId ?? entity.id` for robustness.

**7. No tests**
New route + sync script with entity-type-specific logic but zero tests.

### Minor
- `as CoverageScanRow` cast instead of Drizzle's inferred types
- CHECK constraint in migration but not Drizzle schema
- `loadFactBase()` silently returns empty on error (should hard-fail or warn prominently)
- Compressed code style hurts readability

OAGr:
Rebased on main to resolve merge conflict. 🤖 PR Patrol

PR #4126, opened 2026-04-10
feat: add per-field verdict writing to source-check pipeline

OAGr:
## Summary

- Extend `writeInlineVerdicts` to write per-field verdicts alongside row-level verdicts when `sourcing.fieldVerdicts` is present
- Add `FieldVerdictSchema` and `fieldVerdicts` field to `InlineSourcingSchema` (Zod schema)
- Add `fetchFieldVerdicts()` function to `source-check-join.ts` for reading per-field verdicts grouped by record ID
- Track `fieldVerdictsWritten` count in sync factory `SyncResponse`

## Design decisions

- Evidence rows remain row-level only (field_name = NULL) because the evidence dedup index (`idx_sce_dedup`) does not include `field_name`. Per-field evidence can be added in a follow-up after the index is extended.
- Backward compatible: existing sync routes only access `.written` from the return value, which is unchanged.
- No existing sync routes need modification in this PR — they will start writing per-field verdicts when they are updated to pass `fieldVerdicts` data.

## Test plan

- [x] 8 new tests for `writeInlineVerdicts`: row-level only, with field verdicts, mixed records, empty fieldVerdicts, upsert, evidence isolation
- [x] 3 new tests for `fetchFieldVerdicts`: empty input, grouped results, NULL field_name exclusion
- [x] Existing sync-factory tests pass (13 tests)
- [x] Full test suite passes (5334 tests, 0 failures)
- [x] TypeScript compiles cleanly (wiki-server + web)

Fixes QUA-203 (partial — step 1 of 3)

Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added support for per-field source verdicts, enabling sourcing confidence and verdict information to be assigned to individual data fields in addition to record-level verdicts
  * Enhanced verdict write response to separately track the count of field-level verdicts written

* **Tests**
  * Added comprehensive test coverage for field-level verdict retrieval and write operations

### DIFF
--- a/apps/wiki-server/src/__tests__/source-check-join.test.ts
+++ b/apps/wiki-server/src/__tests__/source-check-join.test.ts
@@ -48,13 +48,17 @@ const { fetchFieldVerdicts } = await import(
   "../routes/shared/source-check-join.js"
 );
 
+// Get the Drizzle instance from the mock to pass as db parameter
+const dbModule = await import("../db.js");
+const db = (dbModule as any).getDrizzleDb();
+
 // ---------------------------------------------------------------------------
 
 describe("fetchFieldVerdicts", () => {
   beforeEach(resetStores);
 
   it("returns empty map for empty recordIds", async () => {
-    const result = await fetchFieldVerdicts("grant", []);
+    const result = await fetchFieldVerdicts(db, "grant", []);
     expect(result.size).toBe(0);
   });
 
@@ -94,7 +98,7 @@ describe("fetchFieldVerdicts", () => {
       },
     ];
 
-    const result = await fetchFieldVerdicts("grant", ["g1", "g2"]);
+    const result = await fetchFieldVerdicts(db, "grant", ["g1", "g2"]);
 
     expect(result.size).toBe(2);
 
@@ -126,7 +130,7 @@ describe("fetchFieldVerdicts", () => {
       },
     ];
 
-    const result = await fetchFieldVerdicts("grant", ["g1"]);
+    const result = await fetchFieldVerdicts(db, "grant", ["g1"]);
     expect(result.size).toBe(0);
   });
 });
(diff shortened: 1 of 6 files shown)


OAGr:
## Code Review Findings

### Significant

**`fetchFieldVerdicts()` calls `getDrizzleDb()` directly — bypasses transaction boundaries**

If the caller reads verdicts within the same transaction that wrote them, this function uses a different connection and may not see uncommitted writes. PR #4131 has a better version that accepts `db` as a parameter. When merging, #4131's implementation should replace this one.

### Minor
- Sequential per-field INSERT statements (10 round-trips for 10 fields) — consider batching in a follow-up
- `FieldVerdictSchema` accepts any string as field name — consider `z.string().regex(/^[a-z_][a-z0-9_]*$/).max(63)`
- Test mock dispatch relies on `params.length >= 7` heuristic (fragile)

OAGr:
**Merge ordering note**: #4131 has a different `fetchFieldVerdicts` signature that uses `sqlInList()` for parameterized array queries. #4131's version should be considered canonical.

**Recommended merge order**: Merge #4126 first, then #4131 replaces the `fetchFieldVerdicts` implementation. #4131 was written against the old (pre-db-parameter) signature, so it will need a trivial update to keep the `db` parameter after this PR lands.

OAGr:
Addressed CodeRabbit Minor finding in commit `036049cb`:

The test mock's `dispatch()` now filters by `record_type` and `record_ids` from the bound params (not just `field_name IS NOT NULL`). Added 2 out-of-scope rows — one with a different `record_type` (`investment`) and one with an unrequested `record_id` (`g99`) — to verify the filter actually excludes them. 3/3 tests pass. @coderabbitai please re-review. 🤖 PR Patrol

OAGr:
⚠️ Rebase conflict — 5 conflict regions across source-check-join.ts and its test file, caused by interaction with merged PR #4131 (read path). This needs careful manual resolution to combine the write and read paths. Flagging for author attention. 🤖 PR Patrol

PR #4184, opened 2026-04-12
fix: regenerate sourcing lint baseline (unblock main)

OAGr:
Main CI is red — sourcing lint baseline went from 7 to 23 after PR #4176 merged. Regenerated baseline. See QUA-238 for the proper fix.


OAGr:
Closing — going to fix the actual issue (migrate client calls from /api/source-checks/ to /api/sourcing/) instead of just bumping the baseline. Per user feedback.

PR #4188, opened 2026-04-12
feat: strengthen temporal mismatch guideline (source updated since claim)

OAGr:
## Summary

- Add explicit guideline for "source has been updated since claim was made" to the source-check prompt
- Resolves a common false positive: historical timeseries facts whose source URL points to a frequently-updated page (Sacra, Crunchbase, news sites, company about pages)

**Example that triggered this fix**:
- Claim: Anthropic revenue = \$0.1B as of 2023-12
- Source: sacra.com/c/anthropic now shows \$19B as of 2026
- Old verdict: \"contradicted\" (190x difference)
- Correct verdict: \"outdated\" (source updated with newer data)

This pattern accounts for several of the 18 FactBase fact contradictions — the claims are historically correct but the source pages have been updated.

## Test plan
- [x] No code changes, just prompt text
- [ ] Re-verify affected facts after merge to see if verdicts shift contradicted → outdated

Fixes QUA-246


OAGr:
We just changed /source-check to /sourcing, and now this is changing it back? That seems confusing. My guess is that we should stay with /sourcing. we need to really choose.

PR #4195, opened 2026-04-12
Fix raw entity IDs exposed in database/data pages

OAGr:
## Summary

Fix raw entity stableIds (`sid_mK9pX3rQ7n`) leaking into the UI on `/organizations/:slug/data?tab=database` and similar entity profile database views. Resolves the issue surfaced in https://www.longtermwiki.com/organizations/anthropic/data where raw IDs appeared in the Entity and Subject columns of the facts, entity_resources, and things tables.

## Root cause

`"entity_id".endsWith("_entity_id")` is **false** because the suffix `_entity_id` (10 chars) is longer than the column name `entity_id` (9 chars). The frontend's `isEntityRef` check matched compound names like `person_entity_id` but missed bare `entity_id`, so these columns never resolved stableIds to display names.

Additionally, the `subject` column in the `facts` table (which references `entities.stableId`) was not in the entity-ref detection list at all. Same for `org_id` in `funding_programs` and `model_id` in `benchmark_results`.

## Key changes

**Detection fix** — `apps/web/src/app/internal/entity-profile/entity-ref-columns.ts` (new)
- Extract `isEntityRefColumn()` and `isEntityRefKey()` into a testable module
- Add explicit matches for `entity_id`, `subject`, `parent_thing_id`, `org_id`, `model_id`
- Backend counterpart updated in `apps/wiki-server/src/routes/tablebase/entity-profile.ts`

**UI cleanup** — `apps/web/src/app/internal/entity-profile/entity-profile-viewer.tsx`
- Hide redundant `entity_id` column in `facts` and `entityResources` sections (always the entity being viewed)
- Hide internal IDs in `things` section (`source_id`, `source_table`, `parent_thing_id`)
- Add column header overrides: `subject` → "Subject Entity", `parent_thing_id` → "Parent"
- Remove duplicate `DisplayNameEntry` interface declaration

**Test coverage** (138 new tests across 3 layers)
- **Unit tests** (`entity-ref-columns.test.ts`) — 65 tests for positive, negative, and `endsWith` regression cases
- **Schema registry coverage** (`entity-profile-ref-coverage.test.ts`) — 69 tests enumerating all known entity FK columns per entity-profile section with their disposition; this layer caught the `org_id` and `model_id` gaps
- **Playwright e2e** (`no-raw-ids.spec.ts`) — new test suites for `/organizations/*/data?tab=database`, `/people/*/data?tab=database`, and structured data pages; added `f_` FactBase ID pattern; extracted shared `collectTextAcrossTabs()` helper; fixed bare-catch skips to only skip on load failures

## Test plan

- [x] Unit tests — 65/65 pass
- [x] Schema registry coverage — 69/69 pass
- [x] Typecheck (web + wiki-server) — exit 0
- [x] `/agent-review-pr` executed — 8 findings, 3 fixed (duplicate interface, bare catches, tab-scan dedup), 3 documented (cross-package boundary for shared helpers), 2 dismissed (LOW confidence)
- [x] `/simplify` pass — tab-scanning dedup, trivial tests removed
- [ ] Playwright e2e against production after deploy: `cd apps/web && PLAYWRIGHT_BASE_URL=https://www.longtermwiki.com npx playwright test e2e/no-raw-ids.spec.ts`
- [ ] Manually verify `/organizations/anthropic/data?tab=database` shows "Anthropic" instead of `sid_mK9pX3rQ7n` in Entity column

## Known tech debt (out of scope)

The detection logic exists in three places (canonical `entity-ref-columns.ts` in web, inline in `entity-profile.ts` server route, copy in wiki-server test file). Cross-package boundary prevents direct imports. A follow-up should move these helpers to `@longterm-wiki/id-utils` for single-source-of-truth. All three copies are annotated with `SYNC` comments pointing to the canonical source.

### DIFF
--- a/apps/web/e2e/no-raw-ids.spec.ts
+++ b/apps/web/e2e/no-raw-ids.spec.ts
@@ -24,6 +24,11 @@ const RAW_ID_PATTERNS = [
     pattern: /\bsid_[A-Za-z0-9]{3,}\b/,
   },
   {
+    name: "f_ FactBase ID",
+    // FactBase fact IDs like f_0PxEB2yv45 — internal identifiers
+    pattern: /\bf_[A-Za-z0-9]{8,}\b/,
+  },
+  {
     name: "[object Object]",
     // React/JS rendering bug where an object is coerced to string
     pattern: /\[object Object\]/,
@@ -82,19 +87,8 @@ function checkTextForRawIds(text: string, url: string, label?: string) {
   }
 }
 
-async function assertNoRawIds(page: Page, url: string) {
-  await loadPage(page, url);
-  const text = await getMainText(page);
-  checkTextForRawIds(text, url);
-}
-
-/**
- * Variant that clicks through tabs before scanning — tab content is
- * lazily rendered and won't be visible without clicking.
- */
-async function assertNoRawIdsWithTabs(page: Page, url: string) {
-  await loadPage(page, url);
-
+/** Collect visible text from all tabs, clicking through each one. */
+async function collectTextAcrossTabs(page: Page): Promise<{ text: string; tabCount: number }> {
   const tabs = page.locator("[role='tab'], button[data-state]");
   const tabCount = await tabs.count();
   const texts: string[] = [];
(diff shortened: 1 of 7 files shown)


OAGr:
Fixed the `/internal/facts` e2e failure: the facts table in `apps/web/src/app/factbase/factbase-facts-table.tsx` was rendering the raw `factId` as link text (truncated at 12 chars, which still matches the new `/\bf_[A-Za-z0-9]{8,}\b/` pattern). Replaced the ID text with a "view →" link; the factId stays in the href and the title attribute for deep links and dev discoverability, but it no longer appears in innerText. Gate passed all 47 checks.

PR #4223, opened 2026-04-12
fix(citations): repair invalid JSON backslash escapes in fix-inaccuracies

OAGr:
## Summary

The `crux w citations fix-inaccuracies` tool produced **0 proposals** in prod for all flagged citations, despite all 49 unit tests passing (QUA-314).

**Root cause**: LLMs mirror MDX-escaped text like `\$100K` verbatim into JSON string values (e.g. `"Awards of ≈\$100K each..."`). `\$` is not a valid JSON escape sequence, so `JSON.parse()` throws and the silent `catch { return []; }` in `parseLLMFixResponse` returned zero proposals. The mocked unit tests never exercised this code path because their test fixtures were already valid JSON.

## Fix

- Add `repairJsonBackslashEscapes()` that doubles backslashes preceding characters which are not valid JSON escape chars (`"\/bfnrtu`), preserving them as literal backslashes.
- Parser now attempts `JSON.parse`, falls back to parsing the repaired string, and emits a `console.warn` (instead of silently returning `[]`) so future parse failures are visible.
- Apply the same repair to `parseSecondOpinionResponse`.
- Add 6 regression tests; 55 total pass.

## Verification (prod)

All three repro cases from QUA-314 now produce proposals:

| Page | Flagged | Before | After |
|---|---|---|---|
| blueprint-biosecurity | 1 | 0 | 1 |
| polymarket | 12 | 0 | 12 |

## Test plan

- [x] `pnpm vitest run crux/citations/fix-inaccuracies.test.ts` — 55/55 pass
- [x] Dry-run against blueprint-biosecurity produces 1 proposal
- [x] Dry-run against polymarket produces 12 proposals

Fixes QUA-314

Unblocks QUA-307.

## Bug Fixes
* Enhanced citation response parsing with automatic recovery mechanism for invalid JSON escape sequences
* Improved fallback handling to ensure reliable citation processing when responses contain formatting issues or parsing errors

### DIFF
--- a/crux/citations/fix-inaccuracies.test.ts
+++ b/crux/citations/fix-inaccuracies.test.ts
@@ -12,6 +12,7 @@ import {
   parseLLMFixResponse,
   applyFixes,
   enrichFromApi,
+  repairJsonBackslashEscapes,
 } from './fix-inaccuracies.ts';
 import type { FlaggedCitation } from './export-dashboard.ts';
 import type { SectionRewrite } from './fix-inaccuracies.ts';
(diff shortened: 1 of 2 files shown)


OAGr:
@coderabbitai Addressed in commit 1ac128d — added type guard before dereferencing parsed JSON. Please re-review.

PR #4228, opened 2026-04-13
QUA-312: URL-quality Phase 2 — shared module + classifier integration + re-ingest temporal fix

OAGr:
## Summary

Phase 2 of the URL-Quality Classification plan ([Discussion #4221](https://github.com/quantified-uncertainty/longterm-wiki/discussions/4221), parent QUA-110/QUA-113). Wires the URL-heuristic classifier from PR #4222 into the live ingest/enrich pipeline so future resources get classified automatically, and fixes the re-ingest skip guard so content-change actually triggers re-classification.

## Motivation

Phase 1 (PR #4222) added a read-only `sourcing-audit-urls` command that flagged ~3,277 stuck verdicts caused by source URLs pointing at homepages. But the inline URL classifier wasn't reachable from the live pipeline, and `resource-enrich.ts` had a skip guard at L98 that unconditionally returned early for `enriched`/`reviewed` resources — making the "re-classify on re-fetch" temporal story a lie.

This PR closes both gaps without spending any LLM budget.

### `crux/lib/sourcing/url-quality.ts` (new, shared module)
Extracted from inline `crux/commands/sourcing-audit-urls.ts`. Exports:
- `classifyByUrl(url)` — pure, network-free heuristic; returns `purpose: 'homepage' | null`, `confidence`, `reasons`
- `normalizeUrlForJoin(url)` — canonical form for evidence↔resource joins
- `extractHost(url)` — domain grouping
- `FLAG_THRESHOLD = 0.7` — audit-flag threshold
- `FAST_PATH_THRESHOLD = 0.8` — "skip LLM" threshold for callers

### `crux/commands/sourcing-audit-urls.ts`
Replaces inline classifier with imports. Re-exports `classifyByUrl`/`normalizeUrlForJoin`/`extractHost` for any external callers (none today, but the re-export keeps the migration risk-free). Also fixed an unused `CONCURRENCY` shadow and a hardcoded evidence-per-record count that was supposed to use the module constant.

### `crux/resource-enrichment/classify.ts` (Haiku batch CLI)
Applies the URL fast-path **before** LLM batch submission. High-confidence homepages get `resource_purpose='homepage'` + `enrichmentStatus='classified'` written directly via `/api/resources/batch`, bypassing the LLM call entirely. **This is where the cost savings actually land** — Haiku batch cost saved on every fast-path hit. The remaining LLM batch only processes resources the heuristic can't decide.

### `crux/lib/job-handlers/resource-enrich.ts` (Sonnet live ingest)
Three changes:
1. **`contentChanged: z.boolean().optional()`** added to `ResourceEnrichParamsSchema`. Optional for backwards compat — older queued jobs without the field still validate.
2. **Skip guard at L98 bypassed when `contentChanged === true`**, so re-ingest with new content actually re-runs enrichment instead of silently no-op'ing on `enriched`/`reviewed` resources.
3. **URL-heuristic post-LLM override** for `resource_purpose`. The LLM still runs (we need summary, key_points, etc.), but the deterministic heuristic wins for clear homepages — matching the plan's "URL heuristic saves classify-step cost, **not** enrich-step cost".

### `crux/lib/job-handlers/resource-ingest.ts`
Passes `contentChanged: true` in the `resource-enrich` job payload when the new content hash differs from `previousContentHash`. **Varies the dedupKey with a content-hash suffix in that case**, so a still-pending enrich job for the same resource doesn't suppress the new signal via `ON CONFLICT (dedup_key) DO NOTHING`. Stable dedupKey when content unchanged or unknown.

## Tests

128 tests pass across the three changed test files:

- **`crux/lib/sourcing/url-quality.test.ts`** — moved from `crux/commands/sourcing-audit-urls.test.ts`. Same 47 adversarial-input tests + 3 new tests pinning down `FAST_PATH_THRESHOLD` semantics.
- **`crux/lib/job-handlers/__tests__/resource-enrich.test.ts`** — 10 new tests:
  - 5 for `contentChanged` skip-guard bypass (omitted, false, true on enriched, true on reviewed, Zod rejection of non-boolean)
  - 5 for URL fast-path override (bare domain wins over LLM, ambiguous URL doesn't override, /about wins, PDF doesn't override, `enrichmentStatus` stays `enriched` not `classified`)
- **`crux/lib/job-handlers/__tests__/resource-ingest.test.ts`** — 3 new tests for `contentChanged` propagation into the enrich job payload (true → field present + varied dedupKey, false → field omitted, no `previousContentHash` → field omitted)

## Verification

- [x] `npx vitest run crux/lib/sourcing/url-quality.test.ts crux/lib/job-handlers/__tests__/resource-enrich.test.ts crux/lib/job-handlers/__tests__/resource-ingest.test.ts` — **128/128 pass**
- [x] `pnpm crux w validate gate --fix` — **48/48 checks pass**
- [x] `pnpm build` — clean
- [x] `npx tsc --noEmit -p crux/tsconfig.json` — **zero new errors in production code I changed** (one pre-existing TS error in `resource-ingest.test.ts:52` from a mock signature on `main`, untouched by this PR)

## Backwards compatibility

- `ResourceEnrichParamsSchema.contentChanged` is optional → older queued job rows still validate
- `sourcing-audit-urls.ts` re-exports the moved helpers → any straggler imports keep working
- The new dedupKey suffix only varies when `contentChanged === true`; the unchanged path retains the old `resource-enrich:{id}` form

## Follow-ups

- **QUA-313** (urgent): backfill `resource_purpose` across ~20k resources + new `sourcing-mark` command + wiki-server `/verdicts` API extension. This is what actually closes QUA-113's data-quality goal.
- The dedupKey collision edge case is mitigated for the `contentChanged` path; broader dedup-key rework is out of scope.

## Test plan

- [x] All affected vitest files pass
- [x] Validation gate clean
- [x] Build clean
- [x] Production TS clean
- [ ] CI green on push

Fixes QUA-312

## Summary by CodeRabbit

* **New Features**
  * Resource enrichment now detects and responds to content changes, triggering re-enrichment when resource content is updated.
  * Implemented fast-path classification for homepage resources using URL structure analysis.
  * Enhanced batch processing for resource classification updates.

* **Tests**
  * Added comprehensive test coverage for content-change detection and URL-based classification behavior.

### DIFF
--- a/crux/commands/sourcing-audit-urls.ts
+++ b/crux/commands/sourcing-audit-urls.ts
@@ -15,11 +15,17 @@
 
 import type { CommandOptions as BaseOptions, CommandResult } from '../lib/command-types.ts';
 import { listVerdicts, getEvidenceByRecord } from '../lib/wiki-server/sourcing-client.ts';
+import {
+  classifyByUrl,
+  normalizeUrlForJoin,
+  extractHost,
+  FLAG_THRESHOLD,
+} from '../lib/sourcing/url-quality.ts';
+
+// Re-export for callers that still import from this file (Phase 1 compat).
+export { classifyByUrl, normalizeUrlForJoin, extractHost };
 
 // ── Module constants ──
-
-/** Confidence threshold: classifyByUrl score ≥ this marks a row as "flagged homepage". */
-const FLAG_THRESHOLD = 0.7;
 
 /** Parallel evidence-fetch requests. Keeps the wiki-server load light. */
 const CONCURRENCY = 8;
@@ -38,27 +44,6 @@ const CLAMP_LIMIT = 2000;
 const TOP_DOMAINS_SHOWN = 30;
 const TOP_SAMPLES_SHOWN = 20;
 const SAMPLE_IDS_PER_DOMAIN = 3;
-
-/** Query parameters that indicate tracking, not data content. */
-const TRACKING_PREFIXES = ['utm_', 'mc_', 'oly_'];
-const TRACKING_EXACT = new Set(['fbclid', 'gclid', 'yclid', 'msclkid', '_ga']);
-
-/** Query parameters that indicate specific content (force non-homepage classification). */
-const DATA_PARAM_KEYS = new Set(['id', 'v', 'q', 'search', 'article', 'item', 'page', 'post']);
-
-/** Known URL shorteners — cannot classify without following the redirect. */
-const SHORTENERS = new Set([
-  'bit.ly', 't.co', 'goo.gl', 'tinyurl.com', 'ow.ly', 'buff.ly',
-  'is.gd', 'shorturl.at', 'rb.gy', 'cutt.ly',
-]);
-
-/** Single-segment paths that act as homepages on most sites. */
-const HOMEPAGE_PATHS = new Set([
-  '/about', '/about-us', '/about_us', '/aboutus',
-  '/contact', '/contact-us', '/contactus',
-  '/home', '/home.html', '/index', '/index.html', '/index.htm',
-  '/welcome', '/main',
-]);
 
 /** True if the caller wants ANSI color codes (TTY and NO_COLOR unset). */
 function useColor(): boolean {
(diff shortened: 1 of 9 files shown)


OAGr:
## Protected Path: `crux/commands/sourcing-audit-urls.ts`

The "Protected Paths Check" is failing because this PR touches a protected file. The change is **a pure refactor** with **no behavioral change** to the audit command:

1. The inline `classifyByUrl` / `normalizeUrlForJoin` / `extractHost` helpers were extracted verbatim into `crux/lib/sourcing/url-quality.ts` (Phase 2 of Discussion #4221, planned in Phase 1's PR #4222).
2. `sourcing-audit-urls.ts` now imports them from the shared module and **re-exports** them so any external callers (none found in the tree, but kept for safety) keep working.
3. Two cleanups inside the same file:
   - Removed a local `CONCURRENCY = 8` shadow that hid the module-level constant.
   - Replaced a hardcoded `limit: 5` literal with the existing `EVIDENCE_PER_RECORD` module constant.

No functional change to the audit output, the URL heuristic behavior, or any CLI flag. The full test suite (`crux/lib/sourcing/url-quality.test.ts` — 50 tests) was moved alongside the code and still passes.

Please add the `gate:rules-ok` label after reviewing.

OAGr:
@coderabbitai Addressed both Major findings:
- `crux/lib/sourcing/url-quality.ts` (383ad8d): reattach `url.search + url.hash` to the inner URL when unwrapping Wayback, so a wrapped `https://example.com/search?q=foo` no longer gets reclassified as `/search`.
- `crux/resource-enrichment/classify.ts` (337682f): on fast-path `/api/resources/batch` write failure, return `exitCode: 1` instead of swallowing the error.

Please re-review.

Note: "Protected Paths Check" is still red — it's waiting on a human to add the `gate:rules-ok` label after review.

OAGr:
## Self-review findings addressed (e177a4771)

Ran `/agent-review-pr` hostile-reviewer pass on the diff. 16 findings total; pushed a follow-up commit addressing them + rebased on your two interim commits (383ad8d78, 337682f35).

### HIGH fixes

1. **HIGH-1** — `classify.ts` was calling raw `apiRequest('POST', '/api/resources/batch', …)` instead of the typed `upsertResourceBatch()` client. Switched to the typed call, so `UpsertResourceSchema` now validates payloads at compile time. (rule violation: `code-review-guidelines.md` → "API callers must use typed wiki-server client functions")

2. **HIGH-2** — partial-failure handling in `classify.ts` fast-path. Your commit 337682f35 already landed the fail-fast behavior — I preserved that (return `exitCode: 1` on first failed batch), kept my typed-client change on top, and added a test (`classify.test.ts: 'fails fast on fast-path batch write failure'`) that pins the contract.

3. **HIGH-3** — `contentChanged` plumbing is end-to-end ready but **no production caller currently passes `previousContentHash`** to `resource-ingest`. Confirmed across 6 call sites. Filed **[QUA-329](https://linear.app/quantifieduncertainty/issue/QUA-329)** to wire up the server-side list-resources API change that completes the temporal story. Not a regression — the skip-guard still works correctly, the bypass just doesn't activate yet.

### MEDIUM fixes

4. Renamed misleading "fast-path" in `resource-enrich.ts` → "URL-heuristic resource_purpose override". The URL override there runs AFTER the LLM call (saves zero cost); the real cost savings are in `classify.ts` where the heuristic runs BEFORE batch submission. Comment + test describe() updated.

5. `dedupKey` content-hash suffix: use the full 16-char hash (64 bits) instead of truncating to 8 (32 bits). Also dropped the dead `&& contentHash` guard — `contentHash` is always non-null inside the enclosing `result.content.length > 0` branch. Test regex updated.

6. **Test coverage** — added `crux/resource-enrichment/classify.test.ts` (new file, 6 tests) for the fast-path:
   - Bucketing: homepage URLs → fast-path, deep paths → LLM batch
   - Dry-run: no writes, no batch submission
   - Fail-fast: first failed chunk aborts with `exitCode: 1`, no LLM submission
   - Fast-path-only: all candidates classified, no LLM batch
   - Skip already-classified resources entirely
   - Shorteners and PDFs route to LLM (not fast-path)

### LOW fixes

7. Misleading test name in `url-quality.test.ts` — `/about` confidence is 0.85 not exactly 0.8; renamed "at exactly FAST_PATH_THRESHOLD" → "above FAST_PATH_THRESHOLD".

8. Added 2 tests in `resource-enrich.test.ts` pinning the invariant that the URL override is NOT applied when the LLM response fails validation or the LLM call throws. A regression that moved the override above the early-return guard would ship silently otherwise.

## Verification

- `npx vitest run` on the 4 changed test files: **136/136 pass** (was 128 pre-review; +6 classify + 2 resource-enrich LLM-failure tests)
- `pnpm crux w validate gate --fix --no-cache`: **41/41 pass**
- `pnpm build`: clean
- `npx tsc --noEmit -p crux/tsconfig.json`: zero new errors in production code

## Deferred (not fixed this round)

- 7 LOW-severity findings that weren't worth the churn: duplicate 'homepage' string literal (no single source of truth for the enum; broad refactor), HOMEPAGE_PATHS doesn't cover `.html`/`.htm`/`.php` variants (known false-negative, won't hurt anyone), `normalizeUrlForJoin` lacks IPv6/IDN tests (edge-case coverage gap), re-exports in `sourcing-audit-urls.ts` are dead weight (leaving for a future cleanup sweep).
- QUA-329 — production-caller wire-up for `previousContentHash`.

Happy to revisit any of these if you'd rather land them here.

PR #4237, opened 2026-04-13
fix(deploy-tasks): SKIP psql verify when psql/DATABASE_URL missing

OAGr:
## Summary

- `crux gh deploy-tasks verify` now reports `skip` (not `fail`) for `psql` tasks when `psql` is missing on PATH or `DATABASE_URL` is unset.
- Skip messages are actionable: they say exactly what's missing and how to fix (run from deploy environment, install psql, export env var).
- Exit code for `--ci` no longer spuriously returns 1 when verify tasks are un-runnable locally.

## Why

Coordinator workstations don't have `psql` installed and don't have `DATABASE_URL` in `.env.base` (DB lives in k8s). Every `/deploy` Step 5 surfaced spurious FAILs for migration tasks on healthy releases, training agents to ignore verify failures — exactly the wrong lesson.

The fix pre-checks psql commands before running them and emits `status: 'skip'` with a dimmed `skipReason` note, instead of letting `bash -c psql ...` fail with exit 127 (command not found) and get classified as a real FAIL.

Fixes QUA-319.

## Changes

- `crux/commands/deploy-tasks.ts`:
  - New exported `checkPsqlRunnable(command, deps?)` helper with injectable probe for tests.
  - Called from `verify()` after the dry-run check and before `runVerifyCommand()` — same point-of-decision pattern as the existing dry-run path.
  - Added `skipReason` field to `VerifyResult` (surfaces in both JSON and human output).
  - `defaultHasPsql()` caches the probe so N psql tasks trigger one `command -v psql` invocation.
- `crux/commands/deploy-tasks.test.ts` (new):
  - 8 unit tests covering non-psql, missing DATABASE_URL (undefined/empty/whitespace), missing psql binary, both present, priority ordering, and word-boundary false-match protection.

## Verification

- `pnpm exec tsc --noEmit -p crux/tsconfig.json` — 161 errors, same as `origin/main` (no new errors introduced).
- `pnpm exec vitest run crux/commands/deploy-tasks.test.ts` — 8/8 passing.
- Ran `pnpm crux gh deploy-tasks verify` on this workstation (no psql installed): previously reported 3 FAILs for migration tasks; now reports 3 SKIPs with the DATABASE_URL message. The remaining PASS/FAIL breakdown reflects real non-psql verify results only.

## Test plan

- [x] `pnpm exec tsc --noEmit -p crux/tsconfig.json` passes (no regressions vs main)
- [x] `pnpm exec vitest run crux/commands/deploy-tasks.test.ts` passes (8/8)
- [x] `pnpm crux gh deploy-tasks verify` on a coordinator machine without psql → SKIP, not FAIL
- [x] `pnpm crux gh deploy-tasks verify --ci` — JSON output includes `skipReason` field
- [x] `--dry-run` still short-circuits to `[dry-run]` skip before the psql check fires

## Release Notes

* **New Features**
  * Deploy tasks now gracefully skip when `psql` is unavailable or required database configuration is missing, with actionable error guidance to help users resolve issues.

* **Tests**
  * Added comprehensive test coverage for deploy task validation logic.

### DIFF
--- a/crux/commands/deploy-tasks.test.ts
+++ b/crux/commands/deploy-tasks.test.ts
@@ -1,0 +1,98 @@
+/**
+ * Tests for deploy-tasks command handlers.
+ *
+ * Covers:
+ * - checkPsqlRunnable: gates psql verify commands when psql or DATABASE_URL
+ *   are missing from the local environment (QUA-319).
+ */
+
+import { describe, it, expect } from 'vitest';
+import { checkPsqlRunnable } from './deploy-tasks.ts';
+
+describe('checkPsqlRunnable', () => {
+  const PSQL_CMD =
+    'psql "$DATABASE_URL" -c "SELECT * FROM drizzle.__drizzle_migrations ORDER BY created_at DESC LIMIT 5;"';
+  const CURL_CMD = 'curl -sf "$WIKI_SERVER_URL/health" | jq .';
+
+  it('returns runnable=true for non-psql commands regardless of env', () => {
+    const result = checkPsqlRunnable(CURL_CMD, {
+      hasPsql: () => false,
+      getDatabaseUrl: () => undefined,
+    });
+    expect(result).toEqual({ runnable: true });
+  });
+
+  it('returns runnable=false with actionable reason when DATABASE_URL is unset', () => {
+    const result = checkPsqlRunnable(PSQL_CMD, {
+      hasPsql: () => true,
+      getDatabaseUrl: () => undefined,
+    });
+    expect(result.runnable).toBe(false);
+    if (!result.runnable) {
+      expect(result.reason).toMatch(/DATABASE_URL is not set/);
+      expect(result.reason).toMatch(/DB access|export DATABASE_URL/);
+    }
+  });
+
+  it('returns runnable=false when DATABASE_URL is empty string', () => {
+    const result = checkPsqlRunnable(PSQL_CMD, {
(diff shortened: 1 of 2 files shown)


OAGr:
would it be better to install psql for all branches/slots?

OAGr:
My take: probably not — at least not until `DATABASE_URL` distribution gets sorted out.

Installing `psql` itself is cheap (one line in slot bootstrap). The real cost is exposing `DATABASE_URL` to every agent slot:

- `DATABASE_URL` currently lives only in k8s, not `.env.base`. Distributing it widens prod-write blast radius across every slot — any compromised agent session, errant test, or wayward `crux` command on N laptops could now hit prod directly.
- The thing `verify` is checking is *post-deploy state*, which is a server-side fact. Asking each laptop to re-derive that fact requires giving each laptop a privileged credential. That's the wrong direction.
- The current PR's `skip` outcome is honest: "this verifier can't run from here, the deploy already ran, trust the smoke test." That's the right shape for a coordinator-laptop tool.

If we want stronger post-deploy verification, the better target is a wiki-server endpoint (or a small in-cluster job) that runs the migration check server-side and exposes the result via a normal authenticated API. Verifier becomes `crux gh deploy-tasks verify --remote` → 200 OK / 4xx, and prod credentials never leave k8s.

So I'd ship this skip-on-missing PR as-is and file a follow-up for the server-side verification endpoint if you want stronger signal.

Issue #1770, opened 2026-03-05
Set LONGTERMWIKI_CONTENT_KEY in groundskeeper K8s deployment


OAGr:
## Problem

PR #1706 fixed the snapshot-retention and github-shadowban-check groundskeeper tasks, but requires LONGTERMWIKI_CONTENT_KEY to be set as an environment variable in the groundskeeper Kubernetes deployment. Without it, the snapshot-retention task cannot authenticate to the content API. TASK_GITHUB_SHADOWBAN_CHECK_USERNAMES is optional and can be left empty.

## Recommended Model

**Haiku** — well-scoped for this model.

## Acceptance Criteria

- [ ] LONGTERMWIKI_CONTENT_KEY is set in K8s deployment
- [ ] snapshot-retention task runs successfully

PR #4261, opened 2026-04-13
fix(QUA-352): cascade hallucination_risk_snapshots.page_id, tighten retention

OAGr:
## Summary

`hallucination_risk_snapshots` has grown to **~3.3M rows / 905 MB** on prod (~47× the ~70K target). Root cause: `page_id` FK had no `ON DELETE` action, so deleting a `wiki_pages` row left orphan snapshots. Each orphan `page_id` then became its own retention partition and accumulated up to `keep` rows that cleanup never reaped — compounded by a silent-failure path when `LONGTERMWIKI_SERVER_API_KEY` is missing and a dry-run `/cleanup` that was stuck at the default 30s `statement_timeout`.

This PR is **Phase 3** of QUA-352 (the durable code fix). Phase 1 (matview refresh) self-resolved on prod; Phase 2 (one-shot orphan DELETE) is the user's manual post-deploy step.

Fixes QUA-352.

## Deliverables

| # | What | Files |
|---|------|-------|
| 1 | **FK cascade migration**, `DROP` + `ADD … NOT VALID` pattern so the ~900 MB table is never held in ACCESS EXCLUSIVE for a scan (prevents repeat of the QUA-302 lock cascade). `VALIDATE CONSTRAINT` is deferred to a manual post-deploy script because it would fail against the existing orphans. FK lookup is restricted to `confrelid = 'wiki_pages'` so an unrelated constraint on page_id would not be dropped by accident. | `apps/wiki-server/src/schema.ts`, `apps/wiki-server/drizzle/0177_hallucination_risk_snapshots_fk_cascade.sql`, `apps/wiki-server/drizzle/meta/_journal.json`, `apps/wiki-server/scripts/validate-hallucination-risk-snapshots-fk.sql` |
| 2 | **Orphan filter in cleanup query** (belt-and-suspenders). `LEFT JOIN wiki_pages wp ON wp.id = hrs.page_id … WHERE (hrs.page_id IS NOT NULL AND wp.id IS NULL) OR …`. Applies to both the dry-run count and the real delete. | `apps/wiki-server/src/routes/wikibase/hallucination-risk.ts` |
| 3 | **Lower retention default 100 → 30**. Dashboards show "latest" + short history; 100 is overkill. | `apps/groundskeeper/src/config.ts` |
| 4 | **Dry-run `/cleanup` now uses `statement_timeout = 120000`**. Previously stuck at 30s, causing 500s on prod. Wrapped both dry-run SELECTs in the same `beginTransaction` + `SET LOCAL` the real cleanup already uses. | `apps/wiki-server/src/routes/wikibase/hallucination-risk.ts` |
| 5 | **Missing API key is no longer silent success**. Returns `success: false` + emits a Discord warning rate-limited to once per day (per process). Preserves the #1770 circuit-breaker compromise (still short-circuits before touching the cleanup endpoints), but the failure is now visible. Module exports `__resetMissingKeyRateLimitForTest` so tests can reset the counter without `vi.resetModules()`. | `apps/groundskeeper/src/tasks/snapshot-retention.ts`, `apps/groundskeeper/src/tasks/snapshot-retention.test.ts` |
| 6 | **Audit entry** querying `/api/hallucination-risk/stats` (which returns `totalSnapshots` from `pg_class.reltuples` — no seq scan). Fails if > 140,000 (2× target). | `.claude/audits.yaml` |

## Why `NOT VALID` + deferred `VALIDATE`

`hallucination_risk_snapshots` is ~900 MB on prod. A naive `ALTER TABLE … DROP CONSTRAINT + ADD CONSTRAINT … ON DELETE CASCADE` would hold ACCESS EXCLUSIVE for the duration of the validation scan — exactly the foot-gun that caused QUA-302's ~12-hour incident on 2026-04-11. This migration instead:

1. Phase 1 (in Drizzle migration 0177): `DROP CONSTRAINT IF EXISTS` (resolved dynamically via `pg_constraint` to handle legacy constraint names from the `page_id_int` swap + backup by the canonical name), then `ADD CONSTRAINT … NOT VALID` — both milliseconds of ACCESS EXCLUSIVE, no row scan. New writes are enforced immediately.
2. Phase 2 (separate manual script `validate-hallucination-risk-snapshots-fk.sql`): runs `VALIDATE CONSTRAINT` under only `SHARE UPDATE EXCLUSIVE`, after the Phase 2 orphan cleanup has removed existing orphans. Includes a pre-flight orphan check with a clear error message if run too early.

This pattern follows `.claude/rules/database-migrations.md` § "ADD CONSTRAINT … NOT VALID".

## Tests

- Wiki-server: **24 hallucination-risk tests pass** (up from 23). New regression test `deletes orphan snapshots whose page_id no longer exists (QUA-352)` directly injects orphan rows into the store and verifies the cleanup path removes them even when retention alone would keep them. The mock dispatcher gained a `computeCleanupPlan` helper that consults the slug→int map as a proxy for the `wiki_pages` JOIN.
- Groundskeeper: **8 snapshot-retention tests pass** (up from 6). New tests: `returns failure and emits Discord warning when no API key is set`, `rate-limits the missing-API-key Discord warning to once per day`, and `re-fires the missing-API-key Discord warning after 24h elapses` (fake-timers, advances past `ONE_DAY_MS` to verify the counter resets).
- Typecheck clean on both wiki-server and groundskeeper (pre-existing crux TS errors are in the baseline).
- All 6029 tests pass. Full `pnpm build` + `pnpm test` + `pnpm crux w validate gate --fix` green (46/46 gate checks, 0 blocking).
- 2 pre-existing failures in `apps/groundskeeper/src/tasks/issue-responder.test.ts` (unrelated to this PR; confirmed by running the same tests on a clean checkout of `main`).

## Review artifacts

Diff was reviewed by a hostile code-review subagent (`/agent-review-pr`). Findings that were fixed in this PR:
- **HIGH** — migration 0177's dynamic FK lookup now restricts `confrelid = 'wiki_pages'::regclass` (previously could match any single-column FK on `page_id`).
- **MEDIUM** — rate-limit test refactored to use an exported `__resetMissingKeyRateLimitForTest` helper instead of fragile `vi.resetModules()` / dynamic re-import.
- **MEDIUM** — added a fake-timers test for the 24h reset path of the missing-API-key rate limiter (previously only back-to-back suppression was tested).
- **LOW** — fixed misleading comment in `validate-hallucination-risk-snapshots-fk.sql` (said "NOT EXISTS guard", now says "`convalidated` check").
- **LOW** — `.claude/audits.yaml` `added_by_pr: 4261` (was null).

Findings that were intentionally **not** addressed here:
- Phase 2 batched-DELETE orphan-cleanup script — explicitly out of scope per the dispatch brief. See deploy checklist below for ordering.
- `citation_accuracy_snapshots` likely has the same orphan-page_id issue — out of scope; should be tracked in a sibling ticket.
- Rewrite of the existing `NOT IN` retention predicate into `NOT EXISTS` / CTE for optimizer friendliness — not a new issue introduced by this PR; scope creep for an incident fix.
- Module-level rate-limit state assumes `replicas=1` for the groundskeeper — documented in the code comment; the groundskeeper is deployed as a singleton today.

## Out of scope (per dispatch brief)

- **Phase 1**: already self-resolved on prod.
- **Phase 2 one-shot DELETE of orphan rows**: the user runs this manually from `coord/` after this PR merges and deploys. See QUA-352 § "Phase 2 — one-shot cleanup".
- **`citation_accuracy_snapshots`**: QUA-352 notes it likely has the same problem. Not fixed here; should be tracked in a sibling ticket.
- **General NOT VALID gate validator**: that's QUA-294 / PR #4253.

## Deploy Checklist

> **⚠ Critical ordering.** The new FK is added as `NOT VALID`, which enforces future writes but does not scan existing rows. The cleanup query's 2-minute `statement_timeout` is not enough to delete 3.3M backlog rows in one shot, so **the first scheduled groundskeeper retention run after deploy will still fail** unless the Phase 2 orphan cleanup has already run. Disable the daily `snapshot-retention` task (or manually pause the groundskeeper CronJob) while Phase 2 is in progress if the daily window is close.

- [ ] **Verify migration 0177 applied cleanly**: check server logs for "Dropped FK hallucination_risk_snapshots.…" and confirm `/api/health` returns `healthy` (not `degraded`).
- [ ] **Confirm the FK is present as NOT VALID**:
  ```sql
  SELECT conname, confdeltype, convalidated
  FROM pg_constraint
  WHERE conname = 'hallucination_risk_snapshots_page_id_wiki_pages_id_fk';
  ```
  Expect `confdeltype = 'c'` (CASCADE) and `convalidated = false`.
- [ ] **Run Phase 2 one-shot cleanup from `coord/`** (not this PR). Use the batched pattern from `.claude/rules/database-migrations.md` § "batched UPDATE for large backfills" so you don't trip the 2-minute timeout on a single statement. Skeleton:
  ```sql
  DO $$
  DECLARE rows_deleted INT;
  BEGIN
    LOOP
      DELETE FROM hallucination_risk_snapshots
      WHERE ctid IN (
        SELECT hrs.ctid
        FROM hallucination_risk_snapshots hrs
        LEFT JOIN wiki_pages wp ON wp.id = hrs.page_id
        WHERE hrs.page_id IS NOT NULL AND wp.id IS NULL
        LIMIT 10000
      );
      GET DIAGNOSTICS rows_deleted = ROW_COUNT;
      RAISE NOTICE 'Deleted % orphan rows', rows_deleted;
      EXIT WHEN rows_deleted = 0;
    END LOOP;
  END $$;
  ```
  Run this from a psql session using `DATABASE_MIGRATION_URL` so the statement timeout is 0. Expect to remove millions of rows.
- [ ] **Run the validate script**:
  ```bash
  psql "$DATABASE_MIGRATION_URL" -f apps/wiki-server/scripts/validate-hallucination-risk-snapshots-fk.sql
  ```
  Expect "FK … validated." The script aborts cleanly if orphans still exist (so it's safe to run again after another cleanup pass).
- [ ] **Confirm row count dropped**:
  ```
  curl -sSf https://wiki-server.k8s.quantifieduncertainty.org/api/hallucination-risk/stats | jq '.totalSnapshots'
  ```
  Expect a value under ~70K (the steady-state target). The new audit (`hallucination-risk-snapshots-rowcount`) fails above 140K.
- [ ] **Re-enable the groundskeeper retention task** (if you paused it) and verify the next scheduled run reports `success: true` in the Groundskeeper Runs dashboard (E926).
- [ ] **Verify `LONGTERMWIKI_SERVER_API_KEY` is present in the groundskeeper pod** (`kubectl -n longterm-wiki exec deploy/longterm-wiki-groundskeeper -- env | grep LONGTERMWIKI_SERVER_API_KEY`) so the retention task actually runs. If missing, you'll now get a Discord alert on the next scheduled run (rate-limited to once every 24h per process).

## Test plan

- [x] `pnpm tsc --noEmit` on wiki-server and groundskeeper — clean
- [x] `npx vitest run apps/wiki-server/src/__tests__/hallucination-risk.test.ts` — 24/24 pass including new orphan regression test
- [x] `npx vitest run apps/groundskeeper/src/tasks/snapshot-retention.test.ts` — 8/8 pass including Discord-alert, once-per-day rate-limit, and 24h-reset fake-timer tests
- [x] `pnpm test` — 6029 passed, 26 skipped, 0 failed
- [x] `pnpm build` — exits 0
- [x] `pnpm crux w validate gate --fix --no-cache` — 46/46 checks pass (4 advisory, 0 blocking)
- [x] `npx tsx crux/validate/validate-drizzle-journal.ts` — exits 0; migration 0177 registered
- [x] `pnpm crux sys audits list` — new `hallucination-risk-snapshots-rowcount` audit parses and shows as "never checked"
- [x] `/agent-review-pr` — adaptive review with hostile subagent diff review; all blocking findings addressed in this PR

- [ ] `migration` Verify migration 0177 applied on server start — `psql "$DATABASE_URL" -c "SELECT * FROM drizzle.__drizzle_migrations ORDER BY created_at DESC LIMIT 5;"`
- [ ] `schema` Drizzle schema changed — verify wiki-server redeployed and schema is in sync — `curl -sf "$WIKI_SERVER_URL/health" | jq .`
- [ ] `manual-migration` Run manual migration script: psql "$DATABASE_MIGRATION_URL" -f apps/wiki-server/scripts/validate-hallucination-risk-snapshots-fk.sql (after Phase 2 orphan cleanup)

### DIFF
--- a/.claude/audits.yaml
+++ b/.claude/audits.yaml
@@ -354,9 +354,9 @@ audits:
     check_type: hybrid
     check_command: |
       curl -sSf --max-time 10 https://wiki-server.k8s.quantifieduncertainty.org/api/hallucination-risk/stats \
-        | jq -r '.totalSnapshots // "missing"' \
+        | jq -er '.totalSnapshots | numbers' \
         | awk '{
-            n=$1+0;
+            n=$1;
             printf "reltuples=%d threshold=140000\n", n;
             if (n > 140000) { printf "FAIL: over threshold (%dx target of 70000)\n", n/70000; exit 1 }
             else { print "PASS: within threshold" }
--- a/.claude/audits.yaml
+++ b/.claude/audits.yaml
@@ -348,6 +348,34 @@ audits:
         result: pass
         checked_by: "manual"
         notes: "haiku=claude-haiku-4-5-20251001, sonnet=claude-sonnet-4-6, opus=claude-opus-4-6. Current per system context."
+  - id: hallucination-risk-snapshots-rowcount
+    category: data-pipeline
+    description: "hallucination_risk_snapshots row count stays under 140,000 (2× the ~70K target). Growth beyond this threshold indicates the retention task or FK cascade has stopped working — historically this has been a prod-stability issue (QUA-352, QUA-302)."
+    check_type: hybrid
+    check_command: |
+      curl -sSf --max-time 10 https://wiki-server.k8s.quantifieduncertainty.org/api/hallucination-risk/stats \
+        | jq -r '.totalSnapshots // "missing"' \
+        | awk '{
+            n=$1+0;
+            printf "reltuples=%d threshold=140000\n", n;
+            if (n > 140000) { printf "FAIL: over threshold (%dx target of 70000)\n", n/70000; exit 1 }
+            else { print "PASS: within threshold" }
+          }'
+    how_to_verify: |
+      The check_command curls the wiki-server /api/hallucination-risk/stats endpoint
+      (which returns totalSnapshots computed from pg_class.reltuples — fast, no seq scan)
+      and fails if the value exceeds 140,000 (2× the 70,000 target).
+      If this fails:
+      1. Check groundskeeper run history for the snapshot-retention task (Groundskeeper Runs dashboard, E926)
+      2. Verify LONGTERMWIKI_SERVER_API_KEY is set in the groundskeeper k8s deployment
+      3. Run the cleanup endpoint manually: DELETE /api/hallucination-risk/cleanup?keep=30
+      4. Confirm the FK has ON DELETE CASCADE: `SELECT confdeltype FROM pg_constraint WHERE conname = 'hallucination_risk_snapshots_page_id_wiki_pages_id_fk'` — should be 'c'
+    frequency: weekly
+    added: "2026-04-12"
+    added_by_pr: null  # QUA-352 PR
+    last_checked: null
+    last_result: null
+    history: []
   - id: audits-system-adoption
     category: process
     description: "This audits system itself is being used — items are checked on schedule, not abandoned"
--- a/apps/groundskeeper/src/config.ts
+++ b/apps/groundskeeper/src/config.ts
@@ -107,7 +107,7 @@ export function loadConfig(): Config {
         enabled: envBool("TASK_SNAPSHOT_RETENTION_ENABLED", true),
         schedule:
           process.env["TASK_SNAPSHOT_RETENTION_SCHEDULE"] ?? "0 3 * * *", // daily at 3am UTC
-        keep: envInt("TASK_SNAPSHOT_RETENTION_KEEP", 100),
+        keep: envInt("TASK_SNAPSHOT_RETENTION_KEEP", 30),
       },
       sessionSweep: {
         enabled: envBool("TASK_SESSION_SWEEP_ENABLED", true),
--- a/apps/groundskeeper/src/config.ts
+++ b/apps/groundskeeper/src/config.ts
@@ -8,7 +8,7 @@ export interface ShadowbanCheckConfig extends TaskConfig {
 }
 
 export interface SnapshotRetentionConfig extends TaskConfig {
-  /** Number of snapshots to keep per page (default: 100). */
+  /** Number of snapshots to keep per page (default: 30). */
   keep: number;
 }
 
--- a/apps/groundskeeper/src/tasks/snapshot-retention.test.ts
+++ b/apps/groundskeeper/src/tasks/snapshot-retention.test.ts
@@ -10,8 +10,15 @@ vi.mock("../logger.js", () => ({
   },
 }));
 
+vi.mock("../notify.js", () => ({
+  sendDiscordNotification: vi.fn().mockResolvedValue(undefined),
+}));
+
 import { snapshotRetention } from "./snapshot-retention.js";
+import { sendDiscordNotification } from "../notify.js";
 import type { Config } from "../config.js";
+
+const sendDiscordNotificationMock = vi.mocked(sendDiscordNotification);
 
 function makeConfig(keep = 100): Config {
   return {
@@ -45,6 +52,7 @@ describe("snapshotRetention", () => {
   beforeEach(() => {
     config = makeConfig();
     process.env["LONGTERMWIKI_SERVER_API_KEY"] = "test-key";
+    sendDiscordNotificationMock.mockClear();
   });
 
   afterEach(() => {
@@ -109,13 +117,49 @@ describe("snapshotRetention", () => {
     expect(result.summary).toContain("citation_accuracy: FAILED (HTTP 500)");
   });
 
-  it("returns success (skipped) when no API key is set", async () => {
+  it("returns failure and emits Discord warning when no API key is set", async () => {
     delete process.env["LONGTERMWIKI_SERVER_API_KEY"];
+    // Make sure we don't actually hit fetch
+    const fetchMock = vi.fn();
+    vi.stubGlobal("fetch", fetchMock);
 
     const result = await snapshotRetention(config);
 
-    expect(result.success).toBe(true);
+    expect(result.success).toBe(false);
     expect(result.summary).toContain("Skipped");
+    expect(result.summary).toContain("LONGTERMWIKI_SERVER_API_KEY");
+    // No cleanup endpoints should be called when the key is missing.
+    expect(fetchMock).not.toHaveBeenCalled();
+    // Discord warning should fire on first missing-key run.
+    expect(sendDiscordNotificationMock).toHaveBeenCalledTimes(1);
+    const message = sendDiscordNotificationMock.mock.calls[0]?.[1] as string;
+    expect(message).toContain("snapshot retention disabled");
+    expect(message).toContain("LONGTERMWIKI_SERVER_API_KEY");
+  });
+
+  it("rate-limits the missing-API-key Discord warning to once per day", async () => {
+    // Use a fresh module instance so the module-level `lastMissingKeyWarnAt`
+    // counter is reset — otherwise the previous test's call has already
+    // tripped the counter inside the same process.
+    vi.resetModules();
+    const freshModule = await import("./snapshot-retention.js");
+    const notifyModule = await import("../notify.js");
+    const freshDiscordMock = vi.mocked(notifyModule.sendDiscordNotification);
+    freshDiscordMock.mockClear();
+
+    delete process.env["LONGTERMWIKI_SERVER_API_KEY"];
+    const fetchMock = vi.fn();
+    vi.stubGlobal("fetch", fetchMock);
+
+    // First call fires a Discord warning.
+    const r1 = await freshModule.snapshotRetention(config);
+    expect(r1.success).toBe(false);
+    expect(freshDiscordMock).toHaveBeenCalledTimes(1);
+
+    // Second call within the same process/day must NOT fire a second warning.
+    const r2 = await freshModule.snapshotRetention(config);
+    expect(r2.success).toBe(false);
+    expect(freshDiscordMock).toHaveBeenCalledTimes(1);
   });
 
   it("still runs second cleanup when first throws", async () => {
--- a/apps/groundskeeper/src/tasks/snapshot-retention.ts
+++ b/apps/groundskeeper/src/tasks/snapshot-retention.ts
@@ -1,7 +1,16 @@
 import type { Config } from "../config.js";
 import { logger as rootLogger } from "../logger.js";
+import { sendDiscordNotification } from "../notify.js";
 
 const logger = rootLogger.child({ task: "snapshot-retention" });
+
+const ONE_DAY_MS = 24 * 60 * 60 * 1000;
+
+// Module-level state so the missing-API-key Discord alert fires at most once
+// per day even if the task is configured to run more often than daily. Resets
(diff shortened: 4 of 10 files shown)


OAGr:
**Context note from coordinator (post-merge framing fix needed)**

Coordinator session ran the diagnostic queries from QUA-352 § "Diagnostic queries" against prod between the time this PR was dispatched and now. **Three of the original issue's hypotheses turned out to be wrong**, which means this PR's description and deploy checklist are slightly mis-framed even though the code itself is fine and worth landing as defensive cleanup. Posting this so reviewers and whoever runs the deploy don't get confused.

## What the prod data actually shows

- **Zero orphan rows.** `LEFT JOIN wiki_pages ... WHERE wp.id IS NULL` returns 0 distinct page_ids and 0 rows. The orphan-page_id hypothesis is falsified.
- **No FK existed on prod before this PR** — `pg_constraint` shows only PK + `chk_hrs_level`. So migration 0177 is creating a brand new FK from scratch (good — fills a real gap), not modifying an existing one. The dynamic `pg_constraint` lookup in the migration handles this correctly: `DROP CONSTRAINT IF EXISTS` is a clean no-op when nothing is there. The `VALIDATE` script's pre-flight orphan check will pass immediately because there are no orphans.
- **Per-page distribution is uniform**: `min=135, p50=5195, p95=5264, p99=5264, max=5264`. Every page is at ~5,200 rows. Not orphan accumulation — uniform overflow across all 774 distinct page_ids.
- **The actual root cause** is that the `/api/hallucination-risk/cleanup` query has timed out at 120,000 ms every single nightly run for the last 9+ days. `groundskeeper_runs` rows for `snapshot-retention` event=`failure`: `Retention (keep=100): hallucination_risk: FAILED (The operation was aborted due to timeout)`, `duration_ms ≈ 120,000`, going back to at least 2026-04-05. Circuit breaker has tripped/recovered multiple times. The window-function + `WHERE id NOT IN (subquery)` anti-join cleanup query cannot fit in 120 s on a 3.8M-row table without a covering index on `(page_id, computed_at DESC)`.

## What this means for this PR

**The code changes are still useful and should land** — every individual change is correct:
- New FK with cascade (still fills a gap; will start enforcing on new deletes)
- Orphan filter in cleanup query (cheap belt-and-suspenders, no harm even with 0 orphans today)
- Dry-run statement_timeout fix (correct)
- Lower retention default 100→30 (correct, helps reduce future load)
- Silent-success → Discord warning (correct)
- Audit entry (correct)

**But this PR alone does NOT fix the cleanup-query timeout.** After this merges and deploys, the 03:00 retention cron will still time out every night, because:
- The cleanup query is unchanged structurally (same `WHERE id NOT IN (window-function subquery)` shape).
- No new index on `(page_id, computed_at DESC)`.
- `statement_timeout` is still 120,000 ms.
- Lowering `keep` from 100 to 30 doesn't help the planner — the window function still has to sort the entire table.

## What's actually needed for resolution

1. **Phase 2 (coordinator is running this now)**: age-based bulk delete to shrink the table below the threshold where the cleanup query can fit in 120 s. The right SQL is **NOT** the orphan-DELETE in the issue body (zero orphans). It's `DELETE FROM hallucination_risk_snapshots WHERE computed_at < now() - interval '7 days'` (will drop ~2.4M rows). Running batched from a temp psql pod against the prod DATABASE_URL.
2. **Sibling Linear issue (will be filed by coordinator after this PR is open)**: the actual durable fix — covering index `(page_id, computed_at DESC)`, rewrite cleanup query off the `WHERE id NOT IN (subquery)` anti-join (use `DELETE ... USING (LIMIT)` or batched), possibly raise `statement_timeout` from 120 s.

## Suggested edit to this PR's deploy checklist

The current Deploy Checklist step "Run the Phase 2 one-shot DELETE of existing orphans (the SQL in QUA-352 § Phase 2)" should be reworded to: "Coordinator has already run the Phase 2 age-based bulk DELETE before this PR merges. Verify `/api/hallucination-risk/stats` shows `totalSnapshots` in the few-hundred-thousand range (not 3.8M), then proceed with VALIDATE."

The `VALIDATE` script's orphan pre-flight check will pass either way (no orphans either before or after Phase 2), so it can be run any time after migration 0177 lands.

Reviewers: this PR is fine to merge as-is. The framing fix above is descriptive, not blocking.

OAGr:
@coderabbitai Addressed all 5 findings in commit 1da01aa:

**Major** — `.claude/audits.yaml:363`: fail-closed on missing/non-numeric `totalSnapshots`. Using `jq -er .totalSnapshots | numbers` so a missing field exits non-zero before awk; dropped the `n=$1+0` coercion that would silently map missing → 0 → PASS.

**Minor** — `apps/groundskeeper/src/config.ts:110`: updated `SnapshotRetentionConfig.keep` docstring from `default: 100` → `default: 30` to match the actual `envInt(TASK_SNAPSHOT_RETENTION_KEEP, 30)` fallback.

**Major** — `apps/groundskeeper/src/tasks/snapshot-retention.ts:120`: swapped `await` for `void` on the missing-API-key Discord notification. `sendDiscordNotification` uses a plain fetch with no timeout, so awaiting it could stall `snapshotRetention()`. The debounce (`lastMissingKeyWarnAt`) is still updated synchronously.

**Major** — `apps/wiki-server/drizzle/0177_hallucination_risk_snapshots_fk_cascade.sql`: `SET lock_timeout = '5s'` before the `DROP / ADD CONSTRAINT NOT VALID` block and `RESET lock_timeout` after. `NOT VALID` only skips the validation scan — both operations still acquire `ACCESS EXCLUSIVE` and would wait indefinitely without `lock_timeout`, recreating the exact QUA-302 deploy stall this migration was written to fix. Nice catch.

**Minor** — `apps/wiki-server/src/routes/wikibase/hallucination-risk.ts` (lines 482 and 524-530): added `id DESC` as a stable tiebreaker to `ROW_NUMBER() OVER (PARTITION BY ... ORDER BY computed_at DESC, id DESC)` in both retention queries.

Please re-review.

OAGr:
**Sibling Linear issue filed for the actual root-cause fix:** [QUA-366 — cleanup query is structurally broken, needs semi-join rewrite + covering index](https://linear.app/quantifieduncertainty/issue/QUA-366/hallucination-risk-snapshots-cleanup-query-is-structurally-broken)

QUA-366 is the durable code fix that complements this PR. **This PR (#4261) does not need to wait on QUA-366** — it's safe to merge as-is. QUA-366 will land as a separate PR after this one merges.

Coordinator session has already executed the QUA-366 rewrite as a one-shot manual DELETE against prod (full audit trail in QUA-352 thread), so the prod table is currently at the target steady state of ~73K rows. PR #4261 + QUA-366 together close out the structural problem.

PR #4273, opened 2026-04-13
QUA-149: Clean up orphaned source-check verdicts

OAGr:
## Summary

Adds `POST /api/sourcing/cleanup-orphans` and a `crux tb sourcing-cleanup-orphans` CLI wrapper. Dashboard queries already filter orphan verdicts out at read time via `LIVE_RECORDS_CTE`, but the raw rows pile up and skew `COUNT(*)` — the "personnel coverage 127%" coverage-matrix bug was downstream of this.

## What it does

- Walks `source_check_verdicts`, `source_check_evidence`, and `sourcing_url_suggestions` for rows whose `(record_type, record_id)` pair doesn't resolve in `LIVE_RECORDS_CTE`.
- Defaults to **dry-run**. Callers must pass `dryRun: false` (or `--apply` on the CLI) to delete.
- Only record_types listed in the CTE are touched. Rows with an unknown record_type are left alone so a silent schema drift can't trigger mass deletions.
- Runs all three deletes in a single transaction.
- Returns per-type counts so dashboards can show what got removed.

## Files

- `apps/wiki-server/src/routes/sourcing/sourcing.ts` — new `POST /cleanup-orphans` handler
- `crux/lib/wiki-server/sourcing-client.ts` — RPC client
- `crux/commands/sourcing-cleanup-orphans.ts` — CLI command
- `crux/crux.mjs`, `crux/lib/groups.ts` — command registration under `tb` group
- `apps/wiki-server/src/__tests__/sourcing-cleanup-orphans.test.ts` — 9 integration tests

## Test plan

- [x] `vitest run src/__tests__/sourcing-cleanup-orphans.test.ts` — 9/9
- [x] Full wiki-server vitest — 1033 pass, 21 skip (baseline)
- [x] `pnpm crux w validate gate --fix` — all 50 checks pass
- [ ] After merge: `WIKI_SERVER_ENV=prod crux tb sourcing-cleanup-orphans` (dry-run on prod)
- [ ] After merge + spot-check: re-run with `--apply` to actually clean up the 1002→724 personnel drift

Fixes QUA-149

## Release Notes

* **New Features**
  * Added orphan cleanup functionality for sourcing records with dry-run preview support
  * New CLI command to execute cleanup operations with optional record type filtering
  * New API endpoint to remove orphaned records from sourcing tables
  * Cleanup operations report deletion counts and breakdown by record type

### DIFF
--- a/apps/wiki-server/src/__tests__/sourcing-cleanup-orphans.test.ts
+++ b/apps/wiki-server/src/__tests__/sourcing-cleanup-orphans.test.ts
@@ -1,0 +1,179 @@
+/**
+ * Integration tests for POST /api/sourcing/cleanup-orphans (QUA-149).
+ *
+ * The handler runs up to 6 SQL statements depending on dryRun:
+ *  - SELECT orphan verdicts
+ *  - SELECT orphan evidence
+ *  - SELECT orphan url-suggestions
+ *  - (non-dryRun only) DELETE orphan verdicts
+ *  - (non-dryRun only) DELETE orphan evidence
+ *  - (non-dryRun only) DELETE orphan url-suggestions
+ *
+ * The mock dispatcher distinguishes these by query substring and returns
+ * canned rows so we can assert on the counts the handler reports.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { Hono } from "hono";
+import { type SqlDispatcher, mockDbModule, postJson } from "./test-utils";
+
+let capturedQueries: string[] = [];
+
+// Canned row sets per table. Tests can mutate these in beforeEach.
+let verdictRows: Array<{ record_type: string; record_id: string; field_name: string | null }> = [];
+let evidenceRows: Array<{ id: number; record_type: string }> = [];
+let suggestionRows: Array<{ id: number; record_type: string }> = [];
+
+const dispatch: SqlDispatcher = (query, _params) => {
+  capturedQueries.push(query);
+  // Route the query to the right canned result based on which table it reads/writes.
+  // DELETE queries also match on table name.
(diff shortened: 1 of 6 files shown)


OAGr:
## Review fix applied

Addresses adversarial review findings on this PR:

**Blocker — count accuracy**: non-dry-run now uses `DELETE ... RETURNING record_type` inside the transaction. Reported `deleted.verdicts/evidence/suggestions` match actual DB state change with no SELECT/DELETE drift.

**High — memory bound**: dry-run now uses `SELECT record_type, COUNT(*) GROUP BY record_type` instead of fetching every orphan row into Node. A million-orphan set returns O(types) rows, not O(rows).

**High — empty-string recordType**: Zod schema now requires `.min(1)`. Before, `{recordType: ""}` was falsy-coerced to "no filter" so the caller thought they were scoping but actually scanned every type.

**Medium — test false positives**: previous `/record_type\s*=/` assertion matched the static JOIN clauses in every query, so it would've passed even with the filter omitted. Rewrote the dispatcher to capture `(query, params, insideTransaction)` per call and assert on bound params. Also added a begin()/end() spy so regressions that pull a DELETE out of the transaction will fail the "wraps all three DELETEs in a single transaction" test.

**Medium — test gaps**: added assertions for empty-string rejection, non-boolean `dryRun`, DELETE RETURNING shape correctness, GROUP BY usage in dry-run, and `RETURNING record_type` presence on every DELETE. Test count: 9 → 16.

**Low**: help text updated to accurately describe what's transactional (the deletes + their returned counts) vs what's not (dry-run counts, which may drift before --apply runs). Removed the unused `_applied` parameter from `formatOutput`.

PR #4276, opened 2026-04-13
QUA-150: Clamp sourcing orchestrator concurrency to prevent overload

OAGr:
## Summary

On 2026-04-09 a `--concurrency=20` sourcing orchestrator run overwhelmed the wiki-server with concurrent writes, tripped the FortiGuard IPS firewall, and caused cascading timeouts. This PR centralizes concurrency parsing behind a hard cap of 10 with a clear warning when a request is clamped.

## What changed

- `crux/lib/sourcing/concurrency.ts` (new) — `clampSourcingConcurrency()` helper with `DEFAULT_SOURCING_CONCURRENCY=5` and `getMaxSourcingConcurrency()` defaulting to 10.
- `crux/lib/sourcing/concurrency.test.ts` (new) — 23 unit tests covering fallbacks, clamping, env override, string/number/NaN/empty/negative inputs.
- `crux/lib/sourcing/orchestrator.ts` — replaces the inline `parseInt(..) < 1 ? 5 : ..` block with the clamp helper so the same guard applies to both batch and realtime execution paths.
- `crux/commands/sourcing-orchestrate.ts` — help text documents the cap and the `SOURCING_CONCURRENCY_LIMIT` env-var escape hatch.

## Design notes

- **Why 10?** 2× the default of 5, comfortably below the 20 that caused the incident. The max is overridable via `SOURCING_CONCURRENCY_LIMIT=<n>` for future tuning without a code change.
- **Scope**: `suggest-urls.ts` hardcodes `DEFAULT_CONCURRENCY=5` with no CLI flag, so it's not at risk. `citation-auditor.ts` uses its own `pLimit` with default 3. Both left alone.
- **Behavior**: clamped values emit one warning to stderr per invocation — callers see it immediately and the orchestration still proceeds (no crash).

## Test plan

- [x] `vitest run crux/lib/sourcing/` — 452 pass (23 new)
- [x] `pnpm crux w validate gate --fix` — all 50 checks pass
- [x] Manual: `clampSourcingConcurrency(20)` returns 10 and emits the warning verbatim

Fixes QUA-150

### DIFF
--- a/crux/commands/sourcing-orchestrate.ts
+++ b/crux/commands/sourcing-orchestrate.ts
@@ -346,7 +346,11 @@ Options:
   --entity-type=X        Filter by entity type (organization, person, ai-model, ...)
   --entity=X             Filter by entity (org or person stableId)
   --source=X             Source mode: existing | web-search | all (default: existing)
-  --concurrency=N        Number of parallel sourcing (default: 5)
+  --concurrency=N        Number of parallel sourcing (default: 5, max: 10)
+                         Values above 10 are clamped and logged. Set
+                         SOURCING_CONCURRENCY_LIMIT=<n> to raise the cap
+                         for local testing only — higher values have
+                         overloaded the wiki-server in the past (QUA-150).
   --dry-run              Show what would be verified without calling LLM
   --ci                   JSON output
 
--- a/crux/commands/sourcing-orchestrate.ts
+++ b/crux/commands/sourcing-orchestrate.ts
@@ -346,11 +346,13 @@ Options:
   --entity-type=X        Filter by entity type (organization, person, ai-model, ...)
   --entity=X             Filter by entity (org or person stableId)
   --source=X             Source mode: existing | web-search | all (default: existing)
-  --concurrency=N        Number of parallel sourcing (default: 5, max: 10)
-                         Values above 10 are clamped and logged. Set
-                         SOURCING_CONCURRENCY_LIMIT=<n> to raise the cap
-                         for local testing only — higher values have
-                         overloaded the wiki-server in the past (QUA-150).
+  --concurrency=N        Parallel task count (default: 5, hard max: 8).
+                         Values above 8 are silently clamped with a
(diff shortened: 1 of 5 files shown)


OAGr:
## Review fix applied

Addresses adversarial review findings on this PR:

**High — env var escape hatch removed**: `SOURCING_CONCURRENCY_LIMIT` accepted any positive integer, so `SOURCING_CONCURRENCY_LIMIT=1000` would have silently reopened the exact bug the cap exists to prevent. Removed entirely. A future raise should come via a code edit in a follow-up PR, not an env var that ships in operator `.env` files.

**High — cap lowered from 10 to 8**: the review correctly noted each orchestrator task fans out to ~5 wiki-server calls, so cap 10 meant ~50 peak in-flight writes vs ~100 at the QUA-150 incident — only ~2x margin. At 8 the peak is ~40, a 60% reduction from the incident level while still letting operators speed backfills modestly above the default of 5.

**Medium — strict number parsing**: `parseInt("20abc", 10) === 20` silently accepted partial-parse garbage; `parseInt("1e3", 10) === 1` misinterpreted scientific notation. Switched to `Number()` which rejects both and correctly handles hex/scientific when the user does want them.

**Medium — --ci clamp warning no longer silenced**: the clamp message previously only went to stderr. Now it's routed into `OrchestrationSummary.warnings` so `--ci` callers see it in the JSON payload.

**Low — help text, dead NaN check, expanded tests**: help text no longer contradicts itself. Redundant NaN branch removed (Number.isFinite already covers it). Tests: 23 → 25, including new cases for `"20abc"`, `"abc20"`, `"1e1"`, `"0x8"`, whitespace, and an explicit "env var is ignored" regression test.

## Follow-up: QUA-384

A **request-level** semaphore on `crux/lib/wiki-server/client.ts` would be a stronger defense than a task-level cap (the review's point #2 in full). Filed as QUA-384 as a separate follow-up because it's a broader change touching shared infrastructure and affects non-sourcing callers. This PR's cap remains the primary defense until QUA-384 lands.

PR #4297, opened 2026-04-13
fix(QUA-397): mask legacy fact/resource IDs + fix FactBase things title write path

OAGr:
## Summary

Third PR in a week targeting the raw-ID leak class on `/organizations/*/data?tab=database` and entity profile views. Fixes 3 of the 4 bugs from the dispatch brief (4th was a coordinator misdiagnosis, see below) + adds a build-time Playwright gate so the next regression blocks PR merges instead of post-deploy smoke tests.

Fixes QUA-397.

### Bug B — Facts "Fact ID" column leaked legacy 8-char hex IDs

**Root cause:** The existing `CellValue` regex `^f_[A-Za-z0-9]{8,}$` only matched the new `f_`-prefixed format. Pre-migration facts in the prod DB still have bare hex IDs like `e7c42d88`, `f2a06bd3`. They fell through to a plain `<span>` and rendered the raw ID.

**Fix:** Added `isLegacyFactId(value, columnName)` in a new `id-masking.ts` module. Column-gated to `{fact_id, factId}` so the permissive hex regex doesn't produce false-positive links on unrelated columns (git SHAs, content hashes). Legacy-detected values route through the same `/factbase/fact/<id>` link as the new format.

### Bug C — `things.title` contained raw fact IDs

**Root cause:** `apps/wiki-server/src/routes/factbase/facts.ts` (dual-write to `things` table) built titles as `${factLabel} — ${entityName}` where `factLabel = f.label || f.factId`. When `f.label` was missing (~14% of rows), the raw fact ID leaked into both `title` and `description`, visible in the Entity Profile Viewer, `/things/[id]` pages, and search.

**Fix (write path):** New `buildFactThingStrings()` pure helper with fallback chain `label → measure → entityName`. `measure` is a readable kebab-case alternative (`total-funding`, `interpretability-team-size`). If both are missing, title is just the entity name — less differentiated but never leaks an internal ID.

**Fix (render defense):** `stripFactIdPrefix()` in `CellValue` masks any pre-existing bad row (`f_xxx — Entity` / `e7c42d88: value`) until the next full sync backfills them. Belt-and-suspenders so the UI is clean the moment this PR deploys.

### Bug D — Entity Resources "Resource ID" column leaked 16-char hex

**Root cause:** Legacy pre-`sid_` resource stableIds (`1f21fae8ed666710`, `0116b24a50f52f44`) didn't match `isAnySid()` and fell through to plain text.

**Fix:** Column-gated `isLegacyResourceId` (`{resource_id, resourceId, stable_id, stableId}`) renders them as a muted em-dash with the full ID in `title=` for debugging. No detail-page link — there isn't one keyed by these legacy IDs, and the dispatch brief explicitly scopes the resource-ID data backfill out.

### Bonus — First-column row-ID truncation leaked `sid_xxx` substrings

`ProfileSection`'s first-column "ID" cell rendered `idStr.slice(0, 7)`. For a fact thing whose `idStr` starts with `sid_`, this produced the 7-char substring `sid_mK9` — a partial stableId already flagged by the post-deploy `no-raw-ids.spec.ts` (with its `sid_[A-Za-z0-9]{3,}` pattern). Replaced with `#<rowIndex>`; full ID remains in `title=`.

## Systemic fix — build-time raw-ID gate

Per the QUA-397 brief ("N+ related symptoms in a narrow window"), this is the **third** attempt at the same leak class in a week. The post-deploy `no-raw-ids.spec.ts` only fires AFTER a bad image ships. I added the same assertion block to `render-audit.spec.ts`, which runs in the PR-gate suite.

Pages covered (all seven from the brief):
- `/organizations/anthropic` (non-data, for the divisions view)
- `/organizations/{anthropic,deepmind,meta-ai}/data?tab=database`
- `/people/{dario-amodei,sam-altman,demis-hassabis}/data?tab=database`

Patterns: `f_[A-Za-z0-9]{8,}`, `sid_[A-Za-z0-9]{10}`, and `(?=[a-f0-9]*[a-f])[a-f0-9]{16}` / `[a-f0-9]{8}`. The lookahead requires at least one hex letter so pure-numeric strings like `25502026` (year concatenations) don't produce false positives — the first test run surfaced exactly this and I tightened the regex before commit.

## Coordinator hypothesis disagreement — Bug A is NOT a bug

The dispatch brief hypothesized Bug A ("Divisions Status dots empty on `/organizations/anthropic`") was caused by PR #4277 removing the `/api/source-checks` backward-compat alias. **This hypothesis is wrong.**

`divisions-section.tsx` uses `getRecordVerdict(recordType, recordId)` from `@/data/tablebase`, which reads from a build-time `record-verdicts.json` file — **not an HTTP call**. I traced it:

- `apps/web/src/data/tablebase.ts:1092` → reads local JSON, keyed as `${recordType}:${recordId}`
- `apps/web/scripts/lib/wiki-server-data.mjs:568` → fetches from `/api/sourcing/verdicts` (NEW path, already migrated)
- Prod wiki-server API (`curl -H Authorization ... /api/sourcing/verdicts?record_type=division`) returns 6 rows for Anthropic divisions, all with `verdict: "unchecked"`

`recordVerdictToStatus("unchecked")` correctly maps to `"not_run"` → the white/gray empty dot. The display is correct; the data is genuinely unchecked. The title attribute on the rendered dot even says "Not checked / Verdict: unchecked" (verified via playwright).

Per the brief's instruction: **"If the verdict lookup in Bug A turns out to be a different root cause ... stop and report before continuing — the coordinator's hypothesis may be wrong."** Stopped and reported; no code change for Bug A. If the user wants divisions actually source-checked, that's a separate ticket scoped to running source-check jobs against division records.

## Tests

- `apps/web/src/app/internal/entity-profile/__tests__/id-masking.test.ts` — **22 tests** covering `isFactId`, `isLegacyFactId`, `isLegacyResourceId`, `stripFactIdPrefix`. Adversarial inputs: uppercase hex, wrong lengths, non-hex chars, already-clean values, column-gating assertions, pure-numeric-string false positives.
- `apps/wiki-server/src/__tests__/facts-thing-strings.test.ts` — **8 tests** for `buildFactThingStrings` covering each fallback rung and the critical "never leak raw factId" assertion.
- Full `pnpm test` (6187 tests across both apps): all pass.
- Playwright `render-audit.spec.ts` raw-ID gate run against a local dev server with this patch applied: **all 7 pages pass**. Run against unpatched prod, the same test correctly fails on `/organizations/anthropic/data?tab=database` (confirming test coverage — it catches the current leak).

## Deploy tasks

**None that block this PR.** Two follow-up items for the coordinator post-deploy (non-blocking):

- [ ] **Things title backfill** — the write-path fix only takes effect for newly-synced rows. Existing bad rows (`f_xxx — Entity`, `e7c42d88: 1000000000`) will auto-refresh on the next full `sync-facts` invocation; no manual backfill needed. The render-layer `stripFactIdPrefix()` masks them in the meantime.
- [ ] **Legacy fact-ID / resource-ID data backfill** — out of scope per the brief. The render-layer fixes mask them; if we want to eliminate them from the data layer, that's a separate data migration.

## Not in scope

- Backfilling legacy fact IDs with `f_` prefix
- Backfilling legacy resource IDs with `sid_` prefix
- Rewriting `entity-profile-viewer.tsx` (~900 lines) — deserves its own refactor PR
- QUA-396 (workflow secret regression) — separate ticket

## Test plan

- [x] `pnpm test` — 6187 tests pass
- [x] `npx tsc --noEmit` in both `apps/web` and `apps/wiki-server` — clean
- [x] `pnpm crux w validate gate` — all 50 checks pass (3 advisory warnings unrelated to this PR)
- [x] Playwright `render-audit.spec.ts` against local dev server with this patch — 7/7 raw-ID tests pass
- [x] Playwright `render-audit.spec.ts` against prod (unpatched) — correctly fails on anthropic/data tab (confirms coverage)
- [x] Manual reproduction via `curl` on prod before/after: see unit tests + Playwright as the automated version

Fixes QUA-397

## Summary by CodeRabbit

* **Bug Fixes**
  * Masked raw identifiers throughout entity profile views to prevent ID leaks across database and profile pages.
  * Improved title and description rendering by removing raw ID prefixes from display text.

* **Tests**
  * Added automated audit suite to detect and prevent future ID leaks across pages.
  * Extended test coverage for ID masking helpers and fact-to-thing data mapping.

### DIFF
--- a/apps/web/e2e/render-audit.spec.ts
+++ b/apps/web/e2e/render-audit.spec.ts
@@ -169,3 +169,120 @@ test.describe("Render audit — critical data tables", () => {
     expect(text).toMatch(/\$\d+(?:\.\d+)?[BMT]/);
   });
 });
+
+/**
+ * Raw-ID leak gate (QUA-397).
+ *
+ * The post-deploy `no-raw-ids.spec.ts` suite has repeatedly caught raw
+ * FactBase / resource / stableId leaks on `/data?tab=database` pages after
+ * each attempted fix. That test only runs after a bad image is already live.
+ * This block mirrors the assertion inside the PR-gate suite so a future
+ * regression of the same class blocks the merge instead of the deploy.
+ *
+ * Pattern coverage:
+ *  - `f_[A-Za-z0-9]{8,}` — FactBase fact IDs
+ *  - `sid_[A-Za-z0-9]{10}` — 10-char stableIds (truncated sub-10-char prefixes
+ *    still used as internal short display values are tolerated)
+ *  - `\b[a-f0-9]{16}\b` — bare 16-char hex (legacy pre-sid_ resource IDs)
+ *  - `\b[a-f0-9]{8}\b` — bare 8-char hex (legacy pre-f_ fact IDs)
+ *
+ * Pages covered intentionally overlap the no-raw-ids suite so any regression
+ * reappears here first. Sections must be expanded for the `entity-profile-viewer`
+ * rows to hydrate — we click every collapsible section header in <main> before
+ * scanning.
+ */
+const RAW_ID_LEAK_PAGES = [
+  "/organizations/anthropic",
+  "/organizations/anthropic/data?tab=database",
+  "/organizations/deepmind/data?tab=database",
+  "/organizations/meta-ai/data?tab=database",
+  "/people/dario-amodei/data?tab=database",
+  "/people/sam-altman/data?tab=database",
+  "/people/demis-hassabis/data?tab=database",
+];
+
+// Legacy-ID patterns use a lookahead requiring at least one a-f letter so pure
+// numeric strings (years, account numbers, formatted dates like `25502026`)
+// don't produce false positives. Real legacy fact/resource IDs are SHA-like
+// random hex and reliably contain letters.
+const RAW_ID_PATTERNS: ReadonlyArray<{ name: string; pattern: RegExp }> = [
+  { name: "f_ FactBase ID", pattern: /\bf_[A-Za-z0-9]{8,}\b/g },
+  { name: "sid_ stableId (full)", pattern: /\bsid_[A-Za-z0-9]{10}\b/g },
+  {
+    name: "legacy resource ID (16-char hex)",
+    pattern: /\b(?=[a-f0-9]*[a-f])[a-f0-9]{16}\b/g,
+  },
+  {
+    name: "legacy fact ID (8-char hex)",
+    pattern: /\b(?=[a-f0-9]*[a-f])[a-f0-9]{8}\b/g,
+  },
+];
+
+async function expandAllSections(page: Page) {
+  // `ProfileSection` uses `role=button` on collapsible headers.
+  // Use `.evaluate` rather than `.click()` so the click doesn't escape to
+  // nested Links (e.g. "Source checks →") and cause a navigation.
+  await page.evaluate(() => {
+    const buttons = document.querySelectorAll('main [role="button"]');
+    for (const b of buttons) {
+      (b as HTMLElement).click();
+    }
+  });
+  await page.waitForTimeout(800);
+}
+
+function findRawIdLeaks(text: string): Array<{ name: string; samples: string[] }> {
+  const leaks: Array<{ name: string; samples: string[] }> = [];
+  for (const { name, pattern } of RAW_ID_PATTERNS) {
+    const matches = text.match(pattern);
+    if (matches && matches.length > 0) {
+      leaks.push({ name, samples: [...new Set(matches)].slice(0, 5) });
+    }
+  }
+  return leaks;
+}
+
+test.describe("Render audit — no raw IDs on database/detail pages (QUA-397)", () => {
+  for (const url of RAW_ID_LEAK_PAGES) {
+    test(url, async ({ page }) => {
+      await loadPage(page, url);
+      await expandAllSections(page);
+
+      // Also click every tab-trigger so tab content is scanned too.
+      const tabs = page.locator("[role='tab']");
+      const tabCount = await tabs.count();
+      const scanText = async (label: string) => {
+        const text = await getMainText(page);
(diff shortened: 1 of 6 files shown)


OAGr:

OAGr:
Closing in favor of #4296, which was opened ~1 hour earlier by a concurrent Claude session working the same ticket. Both PRs targeted QUA-397; #4296 is continuing to iterate (the a9 session is still live and just added a render-layer sanitizer). Merging both would double-patch `entity-profile-viewer.tsx` and `facts.ts`.

Full coordinator writeup of the duplication (crux linear start doesn't dedup, ./ws list showed stale a9 state, session-context slot identifier reported a16 for a9 at one point, macOS LocalHostName drift from MacBook-Pro-4 to MacBook-Pro-5 confused the timeline) will be captured in a Linear follow-up ticket.

Things from this PR that should still land if not already in #4296:
- `apps/web/e2e/render-audit.spec.ts` build-time raw-ID assertions (the systemic fix per QUA-397's N+ red flag — make sure #4296 has an equivalent or file a follow-up)
- Extracted `id-masking.ts` module with unit tests
- The column-name-gated `LEGACY_FACT_ID_COLUMNS` / `LEGACY_RESOURCE_ID_COLUMNS` approach

The #4296 author may want to review the diff from this PR as reference before final push.

OAGr:
@coderabbitai minor / deferred to PR author: Patrol attempted to fix the 8→8-12 pattern extension (`apps/web/e2e/render-audit.spec.ts:216-218`) but the local gate blocked the push due to an unrelated worktree-env issue. The fix (replace `[a-f0-9]{8}` → `[a-f0-9]{8,12}` and rename label to `"legacy fact ID (8-12 char hex)"`) is trivial and the CodeRabbit concern is valid (verified against `apps/web/src/app/internal/entity-profile/id-masking.ts:21` where `LEGACY_FACT_ID_RE = /^[a-f0-9]{8,12}$/`). PR author can apply it in one edit.

PR #4300, opened 2026-04-13
feat(linear): block duplicate session claims in crux linear start (QUA-406)

OAGr:
## Summary

Closes QUA-406. Adds a dedup pre-check to `crux linear start` so two concurrent sessions can't race to claim the same Linear issue. Previously, `start()` had zero dedup — it unconditionally posted a "starting work" comment regardless of existing claims. On 2026-04-13 this caused slots a9 and a16 on the same machine to independently ship competing PRs for QUA-397 (PR #4296 and PR #4297), burning significant compute before the collision was noticed.

## What changed

- **`crux linear start` pre-check** — before posting anything, look for two signals of an active competing session:
  1. Recent `🤖 Claude Code starting work` comments from a **different slot** in the last 24h, not yet superseded by a `🤖 Claude Code finished work` comment.
  2. Open PRs in the wiki repo whose title or body mentions the Linear ID.
  
  Either signal blocks the call with **exit code 2**. The error lists the detected claim(s)/PR(s) so you know who to talk to. Verified live against QUA-397 — correctly flags 3 prior claims + PR #4296.

- **`--force` override** — bypasses the dedup check entirely. The start comment is annotated with `⚠ Claimed with --force despite existing session(s)` so the duplication is captured in Linear history.

- **`agent-checklist init` propagation** — runs the dedup pre-check BEFORE writing any session state (checklist file, DB row, active-agent registration). On collision, init returns exit 2 and leaves **no local artifact**, so a colliding init can be re-run after the race is resolved without cleanup. `init --force` threads through to `linear start --force`.

- **Same-slot = same session** — the slot is the canonical "this is me" signal (derived from the `a<N>` ancestor of `cwd`). A prior claim with the same slot is treated as init-crash recovery or explicit resume, not a collision.

- **Fail-open on API errors** — both `findActiveClaimsByOthers` and `findOpenPrsMentioningLinearId` return `[]` on exception. A Linear or GitHub outage will not wedge every session on the machine. The rare missed-collision cost is less bad than blocking all starts on a transient API glitch.

- **Rule update** — `.claude/rules/github-issue-tracking.md` now describes the actual dedup behavior instead of the aspirational "prevents multiple sessions" language that was the root cause of the ticket's trust hazard.

## Out of scope (will follow up)

These came up in the QUA-406 investigation but are not in this PR to keep the diff focused:

- `./ws list` cross-reference with open PRs and running `claude` processes per slot. The ticket framed this as "stale git state" but `list` reads live via `execSync` — the real weakness is no correlation with Linear/PR state. Not a dedup issue per se; file a follow-up if the coordinator wants slot visibility into who owns what.
- `crux sys dispatch --linear=QUA-NNN --slot=N` wrapper that enforces a pre-flight dedup check structurally so coordinators can't skip it under context pressure.
- Coordinator-side pre-dispatch hook — a symmetric rule that a coordinator dispatching a slot MUST run `crux linear view` + `gh pr list --search QUA-NNN` first. Complements this PR's server-side check.

## Test plan

- [x] Unit tests for `findActiveClaimsByOthers` and `findOpenPrsMentioningLinearId` (13 tests in `crux/lib/linear/dedup.test.ts`) covering: empty, same-slot same-session, cross-slot collision, 24h staleness window, finish-releases-prior, PR filtering, word-boundary false-positives, and fail-open on both APIs.
- [x] End-to-end tests in `crux/commands/__tests__/linear.test.ts` (+11 new): happy path, different-slot collision, same-slot resumption, stale >24h, finish-supersedes-start, open-PR collision, `--force` bypass, both fail-open paths.
- [x] Integration tests in `crux/commands/agent-checklist.test.ts` (+5 new): hard-fail with exit 2 and no checklist written on collision, normal proceed on no collision, `--force` bypass, fail-open, and skip when `LINEAR_API_KEY` is unset.
- [x] Full crux test suite run: 105/105 relevant tests pass. Pre-existing unrelated failures in `fix-contradicted-facts`, `validate-sourcing-coverage`, and two `ci-*` files are not touched by this diff.
- [x] `pnpm crux w validate gate --fix` — all 51 gate checks pass, no new advisories.
- [x] `pnpm build` — succeeds.
- [x] Live smoke test: ran `pnpm crux linear start QUA-397` against the real ticket — correctly returns exit 2, lists 3 prior claims (a9, a16, a16) and open PR #4296.

### DIFF
--- a/.claude/rules/github-issue-tracking.md
+++ b/.claude/rules/github-issue-tracking.md
@@ -21,6 +21,35 @@ For GitHub: posts a start comment and adds the `agent:working` label.
 
 **Note:** `crux sys agent-checklist init` handles both automatically when `--linear=QUA-NNN` or `--issue=N` is provided. You rarely need to call these manually.
 
+## Dedup check — `crux linear start` refuses competing claims
+
+Both `crux linear start QUA-NNN` and `crux sys agent-checklist init --linear=QUA-NNN` run a **dedup pre-check** before posting anything to Linear. The check looks for two signals that another session is already working on the issue:
+
+1. **Recent start comments** — A `🤖 Claude Code starting work` comment posted in the last 24h from a **different slot** (e.g., a9 while you're on a5), not yet superseded by a `🤖 Claude Code finished work` comment.
+2. **Open PRs** — Any open PR in the wiki repo whose title or body mentions the Linear ID.
+
+Either signal is sufficient to block. On a collision, both commands exit with **code 2** (distinct from 1 for other errors), print the detected claim(s)/PR(s), and refuse to write any local session state (checklist file, DB row, active-agent registration). This means a colliding `init` leaves **no artifact** — you can fix the race and re-run without cleanup.
+
+Re-running from the **same slot** is NOT a collision — it's treated as session resumption (init-crash recovery, etc.). The slot is derived from the `a<N>` ancestor of the current working directory.
+
+Both checks are **fail-open**: if the Linear or GitHub API is unreachable, the dedup helper returns empty and the start proceeds. A transient API glitch should not wedge every session on the machine. The downside is a rare missed collision when both APIs are down simultaneously.
+
+### Bypassing with `--force`
+
+If the other session is genuinely abandoned or you have explicit permission to take over, pass `--force`:
+
+```bash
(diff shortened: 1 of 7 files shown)


OAGr:

PR #4322, opened 2026-04-14
feat(sourcing-types): add branded RecordId<T> + asRecordId() (QUA-423 phase 1)

OAGr:
## Summary

Add a branded `RecordId<T>` type + `asRecordId(type, id)` constructor to `@longterm-wiki/sourcing-types`. Phase 1 of QUA-423 — pure addition, zero call-site churn. Phase 2 (narrowing `getSourcingHref` / `getRecordVerdict` signatures + migrating ~17 callers) is a separate follow-up.

## Why

The QUA-417 bug shipped because `getRecordVerdict(recordType: string, recordId: string)` accepted *any* string. The frontend built a composite React list key (`${ownerEntityId}-${record.key}` = `"sid_ULjDXpSLCI-8NUnVSueLS"`) and passed it in. The verdict store indexes by raw grant PK (`"8NUnVSueLS"`), so every lookup returned null — every dot on every person's funding tab rendered "unchecked" and 404'd on click.

A plain-string API can't catch this. A gate validator could only catch *today's* instance, not the next variation. A branded nominal type catches the whole class at compile time.

## Design

```ts
export type RecordId<T extends RecordType = RecordType> = string & {
  readonly __brand: "RecordId";
  readonly __type: T;
};
```

The only way to get a `RecordId<T>` is via `asRecordId(type, id)`, which does **runtime** guards on obvious misuse:

| Guard | Catches |
|---|---|
| Empty string | Caller forgot a null check |
| Length > 200 | Caller passed a URL or display name |
| `sid_X-<alphanum>` | The QUA-417 composite-key shape |
| `<multi-digit>-<multi-digit>` | Two numeric PKs joined |

Heuristics are deliberately narrow — false negatives on real PKs are preferred over false positives on legitimate slugs (`some-name-2024`, `arxiv-2310-12345` are fine).

Distinct types at the compile level: `RecordId<"grant">` is **not** assignable to `RecordId<"personnel">`. A function typed `(id: RecordId<"grant">) => void` can't be called with a personnel id.

## Non-throwing variant

```ts
isCandidateRecordId(value: unknown): value is string
```

For boundary checks (URL params, API responses) where you want to render a 404 instead of crashing.

## Scope

Types + constructor + tests only. No signature changes in this PR.

| File | Change |
|---|---|
[...]

PR #4320, opened 2026-04-14
feat(sourcing): extract VALID_RECORD_TYPES into @longterm-wiki/sourcing-types (QUA-424)

OAGr:
## Summary

Extract `VALID_RECORD_TYPES`, `SOURCING_EXEMPT_TYPES`, `VALID_SOURCE_CHECK_VERDICTS`, and related helpers from `apps/wiki-server/src/api-types.ts` into a new shared workspace package `@longterm-wiki/sourcing-types`.

## Why

Before this change, the frontend imported these types from the backend via a `@wiki-server/api-types` path alias. That broke the "apps don't depend on each other" invariant and made silent drift possible on rename/move. The dead-link bug cluster (QUA-416/417/418) surfaced the cost — when FE code passed `"entity"`, `"race"`, `"model-release"` etc. to `getSourcingHref`, nothing in the type system caught that those weren't in `VALID_RECORD_TYPES`.

This PR lays the foundation (a stable shared import path) for **QUA-423 branded `RecordId<T>`**, which will make passing a composite key or a dead type a compile error.

## What moved

| Export | Before | After |
|---|---|---|
| `VALID_RECORD_TYPES` | `apps/wiki-server/src/api-types.ts` | `@longterm-wiki/sourcing-types` (re-exported from api-types.ts for backcompat) |
| `SOURCING_EXEMPT_TYPES` | same | same |
| `VALID_SOURCE_CHECK_VERDICTS` | same | same |
| `RecordType`, `SourcingExemptType`, `SourcingVerdict` | same | same |
| `isSourcingExempt()` | same | same |

## What's new (not previously in api-types.ts)

Two helpers that were repeatedly inlined at call sites get a canonical home:

```ts
export function isValidRecordType(recordType: string): recordType is RecordType;
export function isLinkableSourcingType(recordType: string):
  recordType is Exclude<RecordType, SourcingExemptType>;
```

`isLinkableSourcingType` is the single predicate `getSourcingHref` needs to decide whether a link would 404 — same check QUA-416 attempted inline. Both are type guards, so callers get narrowing for free.

## What did NOT move

- Zod schemas (`EvidenceBody`, `VerdictUpsertBody`, etc.) stay in `api-types.ts` next to the request validators that use them. The shared package is pure constants + types + predicates.
- Existing import sites of `@wiki-server/api-types` continue to work unchanged — `api-types.ts` re-exports every name. Migrating FE callers to import directly from the package is a mechanical follow-up.

[...]

PR #4323, opened 2026-04-14
fix(sourcing): runtime guard against composite-key recordIds (QUA-423 phase 2)

OAGr:
## Summary

Wire the QUA-423 phase 1 runtime guard (`isCandidateRecordId`) into the three frontend helpers that the QUA-417 bug funneled through: `getSourcingHref`, `getRecordVerdict`, `getFieldVerdict`. Each now fast-returns when the recordId looks like a composite React key or is malformed.

## Why pragmatic phase 2

My original QUA-423 plan said phase 2 = "narrow signatures to require `RecordId<T>` + migrate the ~17 call sites". Grepping the current codebase: 70 call sites across 41 files. That's a large mechanical migration whose defensive value is equivalent to running the runtime check inside the helpers — at least for the QUA-417 bug class, which is the shipped incident we're defending against.

This PR ships the runtime half. The composite-key bug is caught identically; no call-site churn; rollback is trivial. Compile-time signature narrowing is noted as an optional follow-up if call sites ever stabilize enough to justify the review cost.

## The guard

```tsx
if (!isCandidateRecordId(recordId)) {
  console.warn(`[sourcing] ${fn}: ... see QUA-417`);
  return undefined | null;   // depending on the helper
}
```

Each helper now:

| Helper | Behavior on guard-fail |
|---|---|
| `getSourcingHref` | returns `undefined` → dot renders non-clickable |
| `getRecordVerdict` | returns `null` → caller branches as "no verdict" |
| `getFieldVerdict` | returns `null` → same |

Dev mode logs a pointed warning referencing QUA-417 so future contributors hitting the same bug class get a link to the real fix.

## Signature change cascade

`getSourcingHref` returned `string` before; now `string | undefined`. 7 callers across 2 files that fed the return into `<Link href={...}>` needed handling. Rather than wrapping each call with `?? "/sourcing"` fallbacks, added a companion:

```ts
export function getStoredVerdictHref(recordType, recordId): string
```

For callers rendering rows that already came from `/api/sourcing/verdicts` (where recordType + recordId are valid by construction). Return type is `string`, so `<Link>` accepts it directly. Defense-in-depth: falls back to `/sourcing` index with a dev-mode warning if somehow called with a malformed ID.

## Files

| File | Change |
|---|---|
| `apps/web/src/app/sourcing/sourcing-shared.tsx` | `getSourcingHref` return type → `string \| undefined`; add `getStoredVerdictHref` |
| `apps/web/src/data/tablebase.ts` | `getRecordVerdict`, `getFieldVerdict` guarded |
| `apps/web/src/app/sourcing/sourcing-table.tsx` | Use `getStoredVerdictHref` (4 sites + 1 type narrow) |
| `apps/web/src/app/things/[id]/page.tsx` | Use `getStoredVerdictHref` (3 sites) |
| **New**: `apps/web/src/app/sourcing/__tests__/sourcing-href.test.ts` | 11 tests |
| `apps/web/src/data/__tests__/field-verdicts.test.ts` | +5 tests for the guard |

## Test plan

**New runtime-guard tests** (11 cases):

- [x] Returns URL for valid PKs (`8NUnVSueLS`, `sid_A4XoubikkQ`, `f_mEKUPPFYRg`)
- [x] Returns undefined for the exact QUA-417 shape (`sid_ULjDXpSLCI-8NUnVSueLS`)
- [x] Returns undefined for `sid_X-sid_Y` and `12345-67890` composites
- [x] Returns undefined for empty string
- [x] Does NOT flag legitimate hyphenated IDs (`some-name-2024`, `arxiv-2310-12345`, `1-2`)
- [x] URL-encodes recordType + recordId properly
- [x] Dev-mode warning fires with a QUA-417 reference
- [x] `getStoredVerdictHref` falls back to `/sourcing` (never undefined)
- [x] `getStoredVerdictHref`'s string return type holds at compile time

**`getRecordVerdict` / `getFieldVerdict` guard tests** (+5 cases):

- [x] Returns verdict for valid PKs
- [x] Returns null on QUA-417 composite-key shape + logs warning with "QUA-417"
- [x] Returns null on empty string recordId
- [x] Does NOT flag legitimate hyphenated IDs (no QUA-417 warning)
- [x] `getFieldVerdict` applies the same guard

**Regression suite**:

- [x] `pnpm test` (apps/web) — 1555 pass (86 files)
- [x] `npx tsc --noEmit` clean in apps/web

## Stack

Stacked on:
- [#4320](https://github.com/quantified-uncertainty/longterm-wiki/pull/4320) (QUA-424 shared package)
- [#4322](https://github.com/quantified-uncertainty/longterm-wiki/pull/4322) (QUA-423 phase 1 — branded type + `isCandidateRecordId`)

Rebase onto main after those merge.

Closes QUA-423 phase 2

Refs QUA-417 (specific bug this prevents), QUA-408 Phase 4 category W (runtime assertion at the choke point)

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -18,6 +18,7 @@
     "@dagrejs/dagre": "^1.1.8",
     "@longterm-wiki/factbase": "workspace:*",
     "@longterm-wiki/id-utils": "workspace:*",
+    "@longterm-wiki/sourcing-types": "workspace:*",
     "@quri/squiggle-components": "^0.10.0",
     "@quri/squiggle-lang": "^0.10.0",
     "@radix-ui/react-collapsible": "^1.1.11",
--- a/apps/web/src/app/sourcing/__tests__/sourcing-href.test.ts
+++ b/apps/web/src/app/sourcing/__tests__/sourcing-href.test.ts
@@ -1,0 +1,117 @@
+/**
+ * Tests for getSourcingHref / getStoredVerdictHref runtime guards (QUA-423 phase 2).
+ *
+ * The QUA-417 bug shipped because getSourcingHref and getRecordVerdict
+ * accepted any string as recordId. This suite locks in that the runtime
+ * guard (via isCandidateRecordId) rejects the specific composite-key
+ * shape that caused QUA-417, without regressing real-world PK shapes.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import {
+  getSourcingHref,
+  getStoredVerdictHref,
+} from "../sourcing-shared";
+
+describe("getSourcingHref — QUA-423 phase 2 runtime guard", () => {
+  let warnSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    warnSpy.mockRestore();
+  });
+
+  it("returns a URL for valid record PKs", () => {
+    expect(getSourcingHref("grant", "8NUnVSueLS")).toBe(
+      "/sourcing/grant/8NUnVSueLS",
+    );
+    expect(getSourcingHref("personnel", "sid_A4XoubikkQ")).toBe(
+      "/sourcing/personnel/sid_A4XoubikkQ",
+    );
+    expect(getSourcingHref("fact", "f_mEKUPPFYRg")).toBe(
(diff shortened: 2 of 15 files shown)


OAGr:
Closing alongside #4320 and #4322 — the phase 2 runtime guards (and new `getStoredVerdictHref` helper + caller migrations) will be combined with the phase 1 branded type in a single consolidated PR that imports from `api-types.ts` rather than the collapsed package.

PR #4341, opened 2026-04-14
fix(sync-facts): prune orphan facts in PG when removed from YAML (QUA-462)

OAGr:
## Summary

`sync-facts` was upsert-only — facts removed from YAML stayed in PG forever. After the 2026-04-14 release, prod had **658 orphan facts** (`/health` reported `totalFacts: 2867` vs. `2209` in YAML), which is why PR #4321's "fix" for QUA-447 didn't actually clear the alcohol-co facts from `/organizations/constellation`.

This adds a `POST /api/facts/prune` endpoint and wires it into `sync-facts.ts` so the post-sync prune brings PG into agreement with YAML. The endpoint is **scoped per entity** — facts for entityIds not in the request are never touched, so partial syncs cannot delete cross-entity data.

Closes QUA-462. Should also retroactively close out the alcohol-co orphans from QUA-447 the next time the workflow fires.

## Key changes

- **`POST /api/facts/prune`** (`apps/wiki-server/src/routes/factbase/facts.ts`): takes `{ entries: [{entityId, factIds}] }`, deletes facts whose `entityId` is in the request but whose `factId` isn't in the keep list. SELECT + DELETE run inside a single transaction. The dual-write `things` rows are cleaned up in the same transaction. Response is capped at 1000 ids with a `truncated` flag so the payload can't balloon.
- **Shared `factThingKey()` helper** in the same file used by both `/sync` and `/prune` so the `things.source_id` encoding cannot drift between writer and pruner.
- **`pruneFacts()`** (`crux/wiki-server/sync-facts.ts`): groups SyncFacts by entityId, sends batches of up to 1000 entries, tracks per-batch errors. The CLI exits non-zero if any prune batch failed (a 4xx used to be silently swallowed).
- **`loadAndTransformFacts()`** now also returns `entityIdsWithNoFacts` — entities present in the KB but emitting zero source facts (the `constellation.yaml: facts: []` case). These are passed through to `pruneFacts` so the constellation-style orphans are caught.
- **`PruneFactsResult`** added to `crux/lib/wiki-server/facts.ts` via Hono RPC `InferResponseType`, replacing the hand-written response cast.
- **CLI flag**: `pnpm crux wiki-server sync-facts --skip-prune` for operators.

## How safety is preserved

The hardest part of pruning from a sync that batches is **not deleting cross-entity data when the sync is partial**. Two layers:

1. The endpoint only ever deletes facts where `entity_id IN (request.entityIds)`. Entities not in the batch are invisible to the prune.
2. `main()` only calls `pruneFacts` after `result.errors === 0` (and explicitly skips if `--skip-prune` is set, with a belt-and-suspenders guard).

A failed batch in the middle of the sync exits non-zero **before** prune runs, so the next clean run gets to clean up.

## Test plan

- [x] **Server `/prune` (14 tests)**: keep-list pruning, constellation empty-keep case, cross-entity safety, things dual-write cleanup, idempotency, URL-unsafe IDs, large batch (50 entities × 3 facts), response shape contract, `truncated` cap (1002 → 1000+truncated), malformed input rejection.
- [x] **Client `pruneFacts` / `groupFactsByEntity` (14 tests)**: grouping, empty-fact entities, no-duplicate-merge, cap-exceeded throw, batching, multi-batch accumulation, batch-failure resilience, error counting, response shape.
- [x] **Real-loader regression test**: `loadAndTransformFacts` against the real `packages/factbase/data` dir, asserts `entityIdsWithNoFacts.length >= 1` and validates the entries are non-empty stable IDs.
- [x] `pnpm build` — green
- [x] `pnpm crux w validate gate --fix` — 51/51 blocking checks pass
- [x] `pnpm test` for `apps/wiki-server` — 1126/1126 pass (zero regressions; pre-existing snapshot-resources test failures from local-only file are unrelated)
- [x] Hostile reviewer pass via `/agent-review-pr`: 15 findings → 10 fixed, 4 skipped with reason, 1 downgraded.
- [x] Type check both projects (`apps/wiki-server`, `crux`)

## Deploy

The `sync-entities-facts.yml` workflow auto-triggers on push to `main` when `crux/wiki-server/sync-facts.ts` or `apps/wiki-server/src/routes/factbase/facts.ts` changes — both touched here. The first run after merge will execute the new prune step and should reduce `/health.totalFacts` from ~2867 to ~2209 (matching the YAML count).

## Deploy Checklist

- [ ] `manual` After merge, the `sync-entities-facts.yml` workflow should auto-trigger on the path filter (this PR touches `crux/wiki-server/sync-facts.ts` and `apps/wiki-server/src/routes/factbase/facts.ts`). Verify the run completed successfully and pruned the orphan facts: expect `/health.totalFacts` to drop from ~2867 to ~2209 (YAML count). If the workflow didn't fire, trigger it manually with `gh workflow run sync-entities-facts.yml -R quantified-uncertainty/longterm-wiki` — `gh run list --workflow=sync-entities-facts.yml -R quantified-uncertainty/longterm-wiki -L 3`
- [ ] `manual` Confirm `/organizations/constellation` no longer shows the alcohol-co facts (Founded 1945, HQ Victor) — the QUA-447 followup that motivated this fix — `curl -sf https://www.longtermwiki.com/api/facts/by-entity/sid_jwe4re7Ggo | jq .total`

## Release Notes

* **New Features**
  * Added `/api/facts/prune` endpoint to clean up stale facts from the database
  * Added `--skip-prune` CLI option to optionally skip fact pruning during sync operations
  * Implemented automatic cleanup of orphaned facts and associated data records after sync

* **Tests**
  * Added comprehensive test coverage for fact pruning and sync workflows

### DIFF
--- a/apps/wiki-server/src/__tests__/facts.test.ts
+++ b/apps/wiki-server/src/__tests__/facts.test.ts
@@ -2,14 +2,28 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { Hono } from "hono";
 import { mockDbModule, postJson } from "./test-utils.js";
 
-// ---- In-memory store simulating Postgres facts table ----
+// ---- In-memory stores simulating Postgres facts and things tables ----
 
 let factsStore: Map<string, Record<string, unknown>>;
+let thingsStore: Map<string, Record<string, unknown>>; // key: `${source_table}::${source_id}`
 let nextId: number;
 
 function resetStores() {
   factsStore = new Map();
+  thingsStore = new Map();
   nextId = 1;
+}
+
+function thingsKey(sourceTable: string, sourceId: string) {
+  return `${sourceTable}::${sourceId}`;
+}
+
+/** Inject a things row directly. Used for the prune-cleans-things test. */
+function injectThing(sourceTable: string, sourceId: string) {
+  thingsStore.set(thingsKey(sourceTable, sourceId), {
+    source_table: sourceTable,
+    source_id: sourceId,
+  });
 }
 
 /** Composite key for the unique index */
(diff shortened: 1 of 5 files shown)


OAGr:

PR #4342, opened 2026-04-14
fix(tablebase): 4 sync handler cleanups (QUA-433..436)

OAGr:
## Summary

Four small tablebase-sync handler fixes surfaced by the QUA-414 denormalization audit. All four touch different handlers; bundled because they share reviewer context and each would be a trivial standalone PR.

- **QUA-433** — Add `"entity-resource"` to `VALID_THING_TYPES`. `entity-resources.ts` has been writing `thingType: "entity-resource"` via `upsertThingsInTx` since the things table existed, but the canonical enum omitted it. The value is consistent with other join-table types already in the list (`investment`, `equity-position`, `personnel`, `division-personnel`), so option 1 ("add to enum") is applied rather than option 2 ("remove the things write"). No `thingHref()` case is added — join-table things types return `null` href, matching existing behavior.
- **QUA-434** — Remove the dead `href` field from `research-areas.ts::toThing`. `ThingSyncInput` never accepted `href`, so the field was silently dropped. `thingHref()` computes research-area URLs at read time from `sourceId`, so nothing depended on the dropped value. The optional "tighten `ThingSyncInput`" follow-up in the ticket is deferred — scope creep for this PR.
- **QUA-435** — Wire `parentTitle` through `/api/things/sync`. The Zod schema already accepted it, but the factory `toRow` and `conflictSet` both omitted it, so the column was silently nulled for every row coming in through this route. Fix adds `parentTitle` to both + a regression test that posts a row with `parentTitle` and asserts the column is populated.
- **QUA-436** — Replace hardcoded `$` in `grants.ts` / `funding-rounds.ts` `things.description` composers with a new `formatMoney()` helper (`apps/wiki-server/src/routes/shared/format-currency.ts`) that uses `Intl.NumberFormat` with `style: "currency"`. `grants.ts` now passes `g.currency` through, so EUR/GBP/JPY/etc. render with the correct native symbol. `funding_rounds` **has no currency column** — the schema comment declares USD-only — so the call site passes `"USD"` explicitly and routes through the same helper, which means a future `currency` column migration can drop straight in without touching the call site again. The ticket body's "both tables have a currency column" claim is incorrect re: funding_rounds.

### Why bundled

Each fix is <10 lines of code. Four separate PRs would have been pure reviewer overhead given the shared subsystem (tablebase sync handlers) and the shared audit origin (QUA-414 §6). The tests and the gate run once for the bundle.

## Test plan

- [x] `pnpm --filter wiki-server test` — 1131 passing, including new tests
- [x] New `format-currency.test.ts` — 17 cases covering USD/EUR/GBP/JPY symbols, null/undefined amount, string-encoded amounts, NaN, Infinity, zero, negatives, lowercase code, malformed code fallback
- [x] `things.test.ts` — new regression test asserts `parentTitle` persists via `/api/things/sync`; the 22 pre-existing tests still pass after bumping the mock's `COLS` constant from 10 to 11
- [x] `pnpm crux w validate gate --fix` — 51/51 pass (3 advisory warnings are pre-existing)
- [x] `tsc --noEmit` — clean on wiki-server and apps/web

## Out of scope / noted

- Tightening `ThingSyncInput` to a `satisfies`-strict type (QUA-434 optional) — deferred, not requested.
- Adding `funding_rounds.currency` column (QUA-436 schema gap) — out of scope; the helper call already passes `"USD"` so the migration can drop in without code changes.

Closes QUA-433
Closes QUA-434
Closes QUA-435
Closes QUA-436

## Summary by CodeRabbit

* **New Features**
  * Multi-currency formatting support with international currency symbols (USD, EUR, GBP, JPY).
  * Parent title field added to entity metadata.
  * New "entity-resource" entity type introduced.

* **Improvements**
  * Currency formatting now consistently applied across funding data.

### DIFF
--- a/apps/wiki-server/src/__tests__/format-currency.test.ts
+++ b/apps/wiki-server/src/__tests__/format-currency.test.ts
@@ -1,0 +1,77 @@
+import { describe, it, expect } from 'vitest';
+import { formatMoney } from '../routes/shared/format-currency.js';
+
+describe('formatMoney', () => {
+  it('formats USD with $ symbol', () => {
+    expect(formatMoney(1_000_000, 'USD')).toBe('$1,000,000');
+  });
+
+  it('formats EUR with € symbol', () => {
+    expect(formatMoney(1_000_000, 'EUR')).toBe('€1,000,000');
+  });
+
+  it('formats GBP with £ symbol', () => {
+    expect(formatMoney(500_000, 'GBP')).toBe('£500,000');
+  });
+
+  it('formats JPY with ¥ symbol', () => {
+    expect(formatMoney(100_000_000, 'JPY')).toBe('¥100,000,000');
+  });
+
+  it('lowercases currency code input and still formats correctly', () => {
+    expect(formatMoney(1000, 'eur')).toBe('€1,000');
+  });
+
+  it('defaults to USD when currency is null', () => {
+    expect(formatMoney(500, null)).toBe('$500');
+  });
+
+  it('defaults to USD when currency is undefined', () => {
+    expect(formatMoney(500, undefined)).toBe('$500');
+  });
+
+  it('returns null for null amount', () => {
+    expect(formatMoney(null, 'USD')).toBeNull();
+  });
+
+  it('returns null for undefined amount', () => {
+    expect(formatMoney(undefined, 'USD')).toBeNull();
+  });
+
+  it('accepts string-encoded amounts (Drizzle numeric)', () => {
+    expect(formatMoney('2500000', 'USD')).toBe('$2,500,000');
+  });
+
+  it('returns null for non-numeric strings', () => {
+    expect(formatMoney('not a number', 'USD')).toBeNull();
+  });
+
+  it('returns null for NaN', () => {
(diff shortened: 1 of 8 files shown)


OAGr:

OAGr:
🤖 **PR Patrol** — Fix attempt complete (116s, 20 max turns, model: sonnet).

**Issues detected**: self-authored-feedback

**Result**:
xtjs, test) — running normally, not failed

**No changes needed.** The patrol trigger fired because CodeRabbit posted a comment on a self-authored PR, but the review content itself has nothing to address. The PR is waiting for the remaining 3 CI jobs to complete, after which it should be mergeable.

OAGr:

OAGr:
🤖 **PR Patrol** — Fix attempt complete (23s, 20 max turns, model: sonnet).

**Issues detected**: self-authored-feedback

**Result**:
omment contains no actionable items. The previous patrol cycle (116s ago) already made the same determination.

**No changes needed.** PR #4342 is clean, CI-green, and mergeable. The "self-authored-feedback" flag was a false alarm — the only reviewer comment is CodeRabbit saying everything is fine.

PR #4343, opened 2026-04-14
fix(tablebase): four QUA-414 audit follow-ups (QUA-433/434/435/436)

OAGr:
## Summary

Bundled fixes the [QUA-414 audit](https://github.com/quantified-uncertainty/longterm-wiki/blob/main/docs/audits/things-denormalization-audit.md) explicitly carved out as standalone tickets — not subsumed by Phase 4b-B. Ships ahead of Phase 4b-B.1 ([QUA-470](https://linear.app/quantifieduncertainty/issue/QUA-470)) so the composer dispatch table refactor inherits a clean slate.

## What changes

- **QUA-433**: add `entity-resource` to `VALID_THING_TYPES`. `entity-resources.ts:129` was writing this thing_type but it was missing from the canonical list — any future enum-aware validator would have rejected existing rows.
- **QUA-434**: drop the dead `href` field from `research-areas.ts` `toThing`. `ThingSyncInput` has no `href`, so the value was silently dropped. Pure clean-up.
- **QUA-435**: add `parentTitle` to `/api/things/sync` `toRow` and `conflictSet`. The Zod schema accepted it, but `toRow`/`conflictSet` both omitted it, so any caller posting `parentTitle` had it silently written as NULL. The in-memory mock dispatcher in `things.test.ts` is updated to handle the new column slot (`COLS=11`, was 10) and the `makeThing` default.
- **QUA-436**: introduce a shared `formatMoney(amount, currency)` helper using `Intl.NumberFormat` with currency-aware symbols. `grants.ts` now passes `g.currency` (was hardcoded `$`); `funding-rounds.ts` uses the helper with the USD default since the table has no `currency` column yet (filed as a follow-up below).

## Tests

- 14 new unit tests for `formatMoney` covering USD/EUR/GBP/CNY, fractional rounding, string inputs, null/undefined currency, and the unknown-code fallback path.
- Updated `things.test.ts` mock dispatcher for the new `parentTitle` column slot.
- Full wiki-server test suite (1,127 tests) and apps/web test suite (1,570 tests) pass.
- `pnpm crux w validate gate` ✅ all 50 checks pass.

## Follow-ups (not in this PR)

- Add a `currency` column to the `funding_rounds` table (Drizzle migration) so non-USD rounds can be displayed correctly. The helper is already wired in — only one line changes.
- `apps/wiki-server/src/routes/shared/__tests__/sourcing-enforcement.test.ts` is an orphan: it exists but isn't picked up by the vitest include glob (`src/__tests__/**/*.test.ts`). Same root cause that almost bit this PR — moving my new test to `src/__tests__/` was the workaround. Worth either moving the file or broadening the glob.

## Test plan

- [x] `pnpm test` (apps/wiki-server) — 1127 passed, 21 skipped
- [x] `pnpm test` (apps/web) — 1570 passed
- [x] `pnpm crux w validate gate --fix` — 50/50 passed
- [x] `npx tsc --noEmit` (wiki-server) — clean
- [x] `npx tsc --noEmit` (apps/web) — clean

Fixes QUA-433


OAGr:

OAGr:
Closing as duplicate of #4342, which merged first while this session was running. Both PRs fix QUA-433/434/435/436 with the same approach (small differences: file name `format-money.ts` vs `format-currency.ts`, slightly different null handling). The dedup check on `crux linear start` didn't catch the collision because the parallel session and this one both came up before either had registered against the QUA-43x tickets specifically. Filing a session retrospective in the QUA-470 PR.

The composer dispatch table work that was originally going to stack on top of this PR is now its own PR (QUA-470).

PR #4354, opened 2026-04-14
refactor(tablebase): auto-mount routes from registry (QUA-454)

OAGr:
Fixes QUA-454

## Summary

Part 1 of 3 replacing QUA-146 (Tablebase Route Systematization project). Extracts the ~33 hand-written `.route()` calls for TableBase routes in `apps/wiki-server/src/app.ts` into a new `TABLEBASE_MOUNTS` array in `apps/wiki-server/src/routes/tablebase/mount-registry.ts`. `app.ts` now iterates the array instead of listing routes individually.

## Why

Two real failure modes this eliminates:

1. **Route file exists but never wired into `app.ts`** → silent 404s in production. This has happened before — the fix was always "find the missing `.route()` call." Now the registry test catches it in CI.
2. **Silent drift between `crux/tablebase/table-registry.ts` and the filesystem** — a separate problem the QUA-456 follow-up validator will handle, but this PR makes the wiki-server-side mount table explicit and testable, which is prerequisite to that.

Net **−51 lines in app.ts** (289 → 238); 31 tablebase route mounts consolidated into one typed array + a `for` loop.

### `apps/wiki-server/src/routes/tablebase/mount-registry.ts` (new, 144 lines)

Exports `TABLEBASE_MOUNTS: readonly TablebaseMount[]` where each entry is `{ path: string; route: Hono<any, any, any> }`. The `any` generic parameters are the idiomatic escape hatch for a heterogeneous collection of narrowly-typed Hono routes — the first-attempt typing with `Hono` (default generics) failed TS2322 because routes that use `ResolvedEntityVars` env or method-chained schemas aren't structurally assignable to `Hono<BlankEnv, BlankSchema>`. Also exports `TABLEBASE_MANUAL_MOUNTS` — the list of route filenames that the registry test uses to compute the "on disk but not registered" drift set.

### `apps/wiki-server/src/app.ts`

Replaces ~33 lines of `app.route("/api/foo", fooRoute)` with:

```ts
for (const { path, route } of TABLEBASE_MOUNTS) {
  app.route(path, route);
}
```

Adds 5 manual mounts for the routes deliberately excluded from the registry:

```ts
app.route("/api/ids", idsRoute);
app.route("/api/entities", entitiesRoute);
app.route("/api/things", thingsRoute);
app.route("/api/bluesky", blueskyRoute);
app.route("/api/data-sources", dataSourcesRoute);
```

These match the sync-factory exclusion list in `.claude/rules/tablebase-sync-factory.md` — they have bespoke concerns (slug displacement on entities, cross-base indexing on things, non-standard sync path on bluesky, multi-table writes on data-sources, sequence-based ID allocation on ids).

### `apps/wiki-server/src/__tests__/mount-registry.test.ts` (new, 110 lines, **6 tests**)

The regression gate for QUA-454:

1. **no duplicate mount paths** — trivial sanity check, but cheap insurance
2. **every path starts with `/api/` and is kebab-case** — format lock
3. **every mount has a non-null route** — catches import-typo bugs
4. **registry + manual-mount list cover every route file on disk** — ***the core drift check***. Reads the filesystem, subtracts known non-route files (schemas, utilities, the registry itself), and asserts the set equals registered + manual. If this fails, either a new route file was added without a registry entry OR an entry was left pointing at a deleted file.
5. **registry and manual-mount list are disjoint** — no double-mount
6. **≥25 entries (regression guard)** — catches accidental deletion / mass moves into the manual list

### `crux/validate/validate-tablebase-completeness.ts`

Adds `mount-registry.ts` and `entity-profile-descriptions.ts` to `NON_ROUTE_FILES` so the completeness validator doesn't flag them as malformed routes. `entity-profile-descriptions.ts` is a pre-existing column description overlay for `entity-profile.ts` — not a route — and was silently ignored before; I'm adding it to the exclusion list now to prevent future confusion.

## Excluded routes — full list

The 5 routes mounted manually in `app.ts` (NOT in the registry):

| Route | Path | Reason for manual mount |
|---|---|---|
| `entities` | `/api/entities` | Slug displacement + statement_timeout overrides |
| `ids` | `/api/ids` | Sequence-based ID allocation (`nextval('entity_id_seq')`) |
| `things` | `/api/things` | IT IS the cross-base index — not a TableBase member |
| `bluesky` | `/api/bluesky` | Only has `/sync/:did` (external-API variant), no standard `/sync` |
| `data-sources` | `/api/data-sources` | Multi-table writes per call (resources + tabular_sources + snapshots) |

This list is mirrored in `mount-registry.ts` header, `TABLEBASE_MANUAL_MOUNTS`, and the test's `NON_ROUTE_FILES` set. The sync-factory exclusion list in `.claude/rules/tablebase-sync-factory.md` matches 1:1.

## What's NOT in this PR

- **Crux table-registry auto-mount**: the wiki-server registry is self-contained. QUA-456 (Part 3 of 3) will add a cross-file validator checking that the crux `TABLE_CONFIGS` set matches the wiki-server registry + manual list.
- **`crux tb scaffold <name>`**: QUA-455 (Part 2 of 3) is the generator that creates boilerplate for new entity types.
- **Non-tablebase routes**: sourcing, claims, pages, health, operational routes etc. are untouched. The scope of this PR is strictly the `routes/tablebase/*` directory.

## Test plan

- [x] `pnpm --filter wiki-server test` — **1194 passed** (was 1188 pre-PR; +6 new registry tests), 21 skipped
- [x] `pnpm --filter wiki-server typecheck` — clean
- [x] `pnpm build` — full web + data build succeeds
- [x] `pnpm crux w validate gate` — all 42 blocking gate checks pass
- [x] `npx tsx crux/validate/validate-tablebase-completeness.ts` — no blocking findings (mount-registry.ts correctly excluded)
- [x] Drift injection smoke-test: manually add a fake route file (`apps/wiki-server/src/routes/tablebase/fake-route.ts`) and confirm `mount-registry.test.ts` fails with a readable "route files missing from registry" error — *tested locally before commit*
- [ ] CI green on GitHub

## Follow-up

- **QUA-455**: `crux tb scaffold <name>` generator that creates the route file + registry entry + crux table-registry entry in one step, so adding a new entity type is a one-command operation.
- **QUA-456**: cross-file validator that pairs the wiki-server registry against the crux `TABLE_CONFIGS` registry, catching drift between the CLI's view of tablebase routes and the server's.

## Summary by CodeRabbit

* **Tests**
  * Added regression tests to ensure tablebase route registry consistency with actual route files on disk.

* **Chores**
  * Refactored tablebase route registration system for improved maintainability and centralized route metadata management.

### DIFF
--- a/apps/wiki-server/src/__tests__/mount-registry.test.ts
+++ b/apps/wiki-server/src/__tests__/mount-registry.test.ts
@@ -1,0 +1,110 @@
+/**
+ * Coverage check for QUA-454: the tablebase mount registry must stay in
+ * sync with the filesystem. This test is the regression gate that
+ * prevents a new tablebase route file from being added without also
+ * registering it (the exact "route file exists but never wired into
+ * app.ts" drift QUA-454 exists to fix).
+ */
+
+import { describe, it, expect } from "vitest";
+import { readdirSync } from "fs";
+import { join } from "path";
+import {
+  TABLEBASE_MOUNTS,
+  TABLEBASE_MANUAL_MOUNTS,
+} from "../routes/tablebase/mount-registry.js";
+
+// Route files that live in `routes/tablebase/` but are NOT Hono routes.
+// (Constants modules, internal helpers, the registry file itself, etc.)
+const NON_ROUTE_FILES = new Set<string>([
+  "index.ts",
+  "mount-registry.ts",
+  "sync-factory.ts",
+  "sync-factory.test-d.ts",
+  "sync-factory-flag.ts",
+  "sourcing-schema.ts",
+  "audit-log.ts",
+  "write-inline-verdicts.ts",
+  "entity-profile-descriptions.ts",
+]);
+
+const ROUTES_DIR = join(
+  new URL(".", import.meta.url).pathname,
+  "..",
+  "routes",
+  "tablebase",
+);
+
+function listRouteFiles(): string[] {
+  return readdirSync(ROUTES_DIR)
+    .filter(
+      (f) =>
+        f.endsWith(".ts") &&
+        !f.endsWith(".test.ts") &&
+        !f.endsWith(".test-d.ts") &&
+        !NON_ROUTE_FILES.has(f),
+    )
+    .map((f) => f.replace(/\.ts$/, ""))
+    .sort();
+}
+
+/** Path "/api/foo-bar" → filename "foo-bar". */
+function pathToFilename(mountPath: string): string {
+  return mountPath.replace(/^\/api\//, "");
+}
+
+describe("TABLEBASE_MOUNTS registry (QUA-454)", () => {
+  it("has no duplicate mount paths", () => {
+    const paths = TABLEBASE_MOUNTS.map((m) => m.path);
+    const unique = new Set(paths);
+    expect(paths).toHaveLength(unique.size);
+  });
+
+  it("every path starts with /api/ and is kebab-case", () => {
+    for (const { path } of TABLEBASE_MOUNTS) {
+      expect(path).toMatch(/^\/api\/[a-z][a-z0-9-]*$/);
+    }
+  });
+
+  it("every mount has a non-null route object", () => {
+    for (const { path, route } of TABLEBASE_MOUNTS) {
+      expect(route, `route for ${path}`).toBeDefined();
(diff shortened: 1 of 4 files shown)


OAGr:

PR #4370, opened 2026-04-14
benchmark(things): materialized view feasibility for Phase 4b-B.2a (QUA-476)

OAGr:
## Summary

**This is a benchmark-only PR.** No schema changes, no sync handler changes, no column drops. The 2b dispatch (which actually creates the MV in prod and drops `things.title` / `description` / `parent_title`) is separate and blocked on this review.

Phase 4b-B.2a of QUA-408. Dispatched per the coordinator brief at `/tmp/dispatch-qua-476.md`. Verdict: **GO** with four conditions, all documented in `docs/benchmarks/qua-476/README.md` §5.

Adds `docs/benchmarks/qua-476/` containing:

- `things-search-mv.sql` — candidate materialized view definition (not a migration). UNIQUE index on `id` for `REFRESH CONCURRENTLY`, GIN index on `search_vector`, and the secondary btree indexes mirrored from the live `things` table so `/api/things/{search,list,children,:id}` can switch over without query rewrites.
- `run-benchmark.mts` — read-only benchmark runner. Uses only session-local `pg_temp` TEMP tables against `PRODUCTION_DB_URL`. Never issues `REFRESH`, never creates a persistent object, never writes a row. `application_name = 'qua-476-benchmark'` so the runner is identifiable in `pg_stat_activity`.
- `benchmark-results.json` — raw numbers captured 2026-04-14.
- `README.md` — methodology, measured numbers, answers to the three QUA-476 design questions, methodology gaps, and the go/no-go rationale.

## Headline numbers (38,375 rows, 52 MB table)

| Metric | Result |
|---|---|
| Cold refresh (self-SELECT + PK + GIN, 2a shape) | **3.6 s avg** (range 3.2–4.2s over 3 runs) |
| JOIN cost estimate (single LEFT JOIN to `entities`) | inside the noise floor (measured −135 ms vs self-SELECT because caches were warm when the JOIN ran) |
| Composed-shape refresh estimate (2b target, ~20 source tables) | **5–9 s** cold, **10–18 s** with `REFRESH CONCURRENTLY` |
| MV query latency (TEMP table + GIN, 10 queries × 5 samples) | **parity with baseline** — ±2 ms per query at p50 |
| MV size after build | 34 MB (21 MB heap + 13 MB indexes) |
| Hourly refresh duty cycle | ≈ 0.5% (18 s / 3600 s) — room for ~50× data growth |

## Verdict: GO with 4 conditions

1. **Re-verify `REFRESH CONCURRENTLY` blocking behavior** empirically on a persistent benchmark MV before dropping any columns. Current estimate (≈ 2× non-concurrent ≈ 7.3 s for 2a shape) is from postgres docs, not measurement.
2. **Instrument `last_refresh_time` and `row_count`** on the MV in `/internal/data-quality` in the same PR that creates the MV. Staleness-as-a-signal is the QUA-408 north star.
3. **Wire a refresh job** (cron, scheduled wiki-server task, or groundskeeper entry) in the same PR. An MV with no refresh schedule is worse than no MV.
4. **Re-verify query latency from a stable host** to collapse the ~60 ms run-to-run network noise we observed between the two benchmark runs.

Full rationale in `docs/benchmarks/qua-476/README.md` §5.

## Methodology notes worth flagging

- Two back-to-back runs of the benchmark showed ~60 ms of run-to-run variance on the client → prod query path (baseline p50s were 71–94 ms in the first run, 141–149 ms in the second). The MV p50s were stable across both runs (141–151 ms). The first run looked like "MV is ~65 ms slower" — the second run inverted the picture and showed parity. README §1.5 documents the gotcha so 2b knows to re-measure from a stable host.
- Run 3's GIN index build was 1698 ms vs ~900 ms for runs 1 and 2 — managed-instance noise on DigitalOcean postgres. The avg still fits the refresh budget comfortably.
- `REFRESH MATERIALIZED VIEW CONCURRENTLY` behavior was NOT measured empirically — it requires a persistent MV and the dispatcher brief said to check with the coordinator before creating persistent objects on prod. Flagged as Condition 1 above.

## Scope boundary (explicitly)

Does NOT:
- Create the MV in prod
- Drop any `things` columns
- Modify `upsertThingsInTx` or any of the 22 sync handlers catalogued in QUA-414
- Change `/api/things/search`
- Migrate the QUA-470 composer dispatch table

All of those are Phase 4b-B.2b, which is the separate dispatch that consumes this benchmark.

## Test plan

- [x] Gate passes (`pnpm crux w validate gate --fix`)
- [x] Benchmark runner reproduces from a clean checkout with `PRODUCTION_DB_URL` set
- [x] Script is read-only against prod — only `pg_temp` DDL, no `REFRESH`, no row writes
- [x] Numbers in README match `benchmark-results.json` (cross-checked programmatically)
- [x] SQL columns in `things-search-mv.sql` match the live `things` pgTable definition
- [x] `/agent-review-pr` run against the final state (diff-hash-verified marker)
- [x] `/simplify` pass applied (helpers extracted; review-induced refactor re-verified against prod)

## References

- Parent epic: QUA-408
- Audit this builds on: QUA-414 (`docs/audits/things-denormalization-audit.md`)
- Composer dispatch infrastructure: QUA-470
- Related: QUA-397 (raw ID leak symptom that Phase 4b-B structurally eliminates)

Fixes QUA-476


OAGr:
# Coordinator review — accept with one methodology pushback

Reviewed against `/tmp/dispatch-qua-476.md` (the dispatch brief) and spot-checked against `apps/wiki-server/src/schema.ts` for MV column parity. Posting as a non-blocking comment — the PR is mergeable.

## What I verified

- **Scope discipline**: 4 files, all under `docs/benchmarks/qua-476/`, zero production code touched. Matches the brief's "benchmark only" boundary exactly.
- **Safety of `run-benchmark.mts`**: read-only against prod. Only `pg_temp` TEMP tables (session-local, auto-dropped). `application_name = 'qua-476-benchmark'` for ops identifiability. 120s statement timeout. No `REFRESH`, INSERT, UPDATE, DELETE, or persistent DDL anywhere. I spot-checked every SQL-issuing line.
- **MV SQL candidate** (`things-search-mv.sql`): UNIQUE index on `id` for CONCURRENTLY ✓, GIN on `search_vector` ✓, secondary btree indexes mirror the current `things` indexes so `/api/things/{list,children,:id}` can switch with zero query rewrites ✓, `WITH NO DATA` + documented first-refresh gotcha ✓. The 2b composed shape is sketched as a comment block and explicitly not benchmarked.
- **Design questions**: all three from the QUA-476 body are answered in §3 of README.

## Strengths

Methodologically honest throughout. Every weakness is flagged in the PR body rather than buried:

- §1.4 JOIN-cost measurement is called unreliable by the author
- §1.5 run-to-run network variance (~60ms) is documented with a recommendation to re-measure from a stable host
- `join_vs_self_delta = -135ms` is explicitly called out as noise, not signal
- The decision to NOT create a persistent MV on prod is correctly cited as the dispatcher brief's scope boundary

This level of self-skepticism is the right bar for a benchmark that gates a multi-phase migration.

## The one real pushback — the 2b composed-refresh estimate is the weakest number

The PR's rationale cites "5-15s pessimistic, 10-30s CONCURRENTLY, ~0.8% hourly duty cycle" as the budget-comfort argument for GO. But:

1. **The 5-15s band is extrapolated**, not measured. The benchmark measured a self-SELECT from `things` itself (3.64s avg, range 3.2-4.2s). The 2b composed MV will be a `UNION ALL` across ~20 source tables — a completely different query shape.
2. **The §1.4 JOIN cost measurement was not a representative sample**. Only 6% of rows have a non-NULL `parent_thing_id`, so the self-join only exercised FK-resolution on 6% of the data. The measured `-135ms` delta is a warm-cache artifact, not a real signal.
3. **The honest bounds are wider**: lower bound is 3.4s (self-SELECT only), upper bound could realistically be 30-60s if a 20-table UNION ALL fans out badly on a managed Postgres instance. Nobody has measured this.

The README acknowledges the weakness — Condition 1 says "re-verify on a restore instance." That's the right mitigation. **But the GO verdict's confidence argument rests partially on numbers that weren't actually measured**, which I want on record so phase 2b doesn't treat the 5-15s band as load-bearing.

**Recommend**: Phase 2b's first step must be a real 20-table `UNION ALL` benchmark (at minimum entities, facts, grants, personnel, resources — 5 tables to cover ~80% of source rows per §1.2 breakdown). Abort-and-escalate thresholds: cold refresh > 60s, CONCURRENTLY > 90s. This makes Condition 1 load-bearing rather than "verify later."

QUA-506 will be updated to cite this as an explicit pre-schema-change deliverable.

## Minor notes (not blocking)

- **GIN build variance**: 857 / 961 / 1698 ms — a 2× spread across 3 runs. DigitalOcean noisy-neighbor effect. Doesn't change the budget math but makes the "3.64s avg" less precise than it reads. A 5-run benchmark would have collapsed this.
- **Query latency is network-bound**: baseline and MV p50s are both ~148ms, of which ~130ms is RTT per §1.5. The "parity" claim is robust (both are DB-side fast), but a reader should not interpret 148ms as "the MV is 148ms fast." Condition 4 (re-run from a stable host) closes this.
- **`/api/things/list` and `/children` not benchmarked**. The MV has the right columns and the secondary indexes mirror the current table, so these endpoints should switch cleanly — but nobody measured. A 10-minute ad-hoc p50 check against a persistent MV during QUA-506 would remove the residual uncertainty.
- **`parent_thing_id` gap deferred** (per §3 answer to design Q3). Fine to defer, but QUA-506's dispatch brief should call out the audit §6 finding #7 so whoever picks it up doesn't flag pre-existing state as a regression.

## Acknowledgment of self-discipline

Explicit acknowledgment: the agent chose **not** to create a persistent MV on prod because the dispatcher brief said "check with coordinator first before touching prod." That was the correct call. Hitting a scope boundary and halting (rather than improvising) is exactly how a good dispatch should behave. The 4 GO conditions are the correct mitigation — they convert the scope-boundary gap into explicit phase-2b deliverables.

## Decision

**APPROVE for merge.** The deliverable is an audit artifact + prototype, not production code. CI should pass trivially.

After merge, QUA-506 will be dispatched with an updated ticket body citing all 4 original conditions + the hardened Condition 1 (UNION ALL benchmark as first step, with abort thresholds) as mandatory deliverables. Dispatch brief for QUA-506 will also call out the `parent_thing_id` pre-existing gap as not-a-regression.

Coordinator: a6 session, 2026-04-15.

OAGr:

PR #4494, opened 2026-04-19
feat(validate): add FK-swap double-drop gate validator (QUA-609)

OAGr:
## Summary

Migration 0186 (the QUA-549 pilot) only dropped the Drizzle-generated FK
name `<table>_<col>_resources_id_fk` and missed the postgres-default
`<table>_<col>_fkey`. Prod had `_fkey` surviving from an older code path,
the Step 2 UPDATE tripped that surviving FK, and the deploy halted.

Every later Phase B migration (0187–0197) drops both names — but the
difference between "safe" and "breaks prod" was the PR author's memory
of the post-mortem. This PR promotes that memory to a blocking gate
check.

## How the validator works

`crux/validate/validate-fk-swap-double-drop.ts` scans
`apps/wiki-server/drizzle/*.sql`. For each file, it detects FK-swap
operations against `resources(stable_id)` for any `(table, column)` pair
via two signals:

1. The canonical hex16 → sid_ rewrite:
   `UPDATE <table> [alias] SET <col> = r.stable_id FROM resources r WHERE [<prefix>.]<col> = r.id`
2. The new FK target:
   `ALTER TABLE <table> ... ADD CONSTRAINT ... FOREIGN KEY (<col>) REFERENCES resources(stable_id)`

For every `(table, col)` discovered, it requires either:

- **Two-name pattern** (used by 0188–0197): both
  `DROP CONSTRAINT IF EXISTS "<table>_<col>_resources_id_fk"` and
  `DROP CONSTRAINT IF EXISTS "<table>_<col>_fkey"` clauses present, OR
- **Dynamic-drop pattern** (used by 0187): a DO block that looks up the
  FK by `(table, column, ccu.table='resources', ccu.column='id')` in
  `information_schema` and `EXECUTE format`s an
  `ALTER TABLE "<table>" DROP CONSTRAINT %I`. This is strictly safer
  because it handles any constraint name, so it's accepted as
  equivalent.

Migrations with no FK-swap signals are skipped entirely. Comments are
stripped before all checks so a doc comment mentioning `DROP CONSTRAINT`
or a commented-out UPDATE doesn't trick the validator either way.

## Verification

- 28 unit tests cover compliant patterns (two-name + dynamic),
  fabricated single-drop violations, multi-block migrations, comment
  stripping, aliasless UPDATE, ADD CONSTRAINT detection across unrelated
  ALTER TABLE statements, and the dynamic-drop spoof-resistance check.
- Standalone snapshot test against current `main` returns 0 violations
  (9 FK-swap migrations correctly classified: 0186–0189, 0191–0194,
  0197).
- Wired into `validate-gate.ts::parallelChecks` as blocking. Local gate
  run: 3.7s for this check, all 42 gate checks pass.

## Review

`/agent-review-pr` ran a hostile-reviewer subagent that flagged three
real detection gaps; all fixed before ship:

- `UPDATE_RE` required an alias, missing aliasless `UPDATE "x" SET ...`
  forms (the exact deploy-halt class the validator should catch).
- `ADD_CONSTRAINT_RE` used `[\s\S]*?` between `ALTER TABLE` and
  `ADD CONSTRAINT`, allowing the regex to span across `;` and
  misattribute the FK to an unrelated earlier ALTER TABLE.
- Two silent `catch {}` blocks lacked the explanatory comments required
  by `.claude/rules/error-handling.md`.

`/simplify` then deduplicated the `findSwapTargets` call (was running
twice per FK-swap file) and extracted a `Signal` type alias plus
`drizzleFkName()` / `postgresFkName()` helpers.

## Test plan

- [x] 28 unit tests pass (`npx vitest run validate/validate-fk-swap-double-drop.test.ts`)
- [x] Validator standalone returns 0 violations on main (`npx tsx crux/validate/validate-fk-swap-double-drop.ts`)
- [x] Full gate passes (`pnpm crux w validate gate --fix --no-cache`)
- [x] TypeScript clean (`npx tsc --noEmit -p crux/tsconfig.json` shows no errors in new files)
- [x] Adversarial fuzzing: empty file, huge file, comment-spoofed DROPs, false-positive UPDATE on non-resources column — all handled correctly
- [ ] CI green on push

Closes QUA-609.

## Summary by CodeRabbit

* **New Features**
  * Added FK-swap double-drop validator ensuring database migrations properly drop constraints when rewriting foreign key columns.
  * Recognizes two acceptable constraint drop patterns: explicit `DROP CONSTRAINT` statements or dynamic `information_schema` lookups.
  * Reports detailed violations listing missing constraint drops and affected migration targets.

### DIFF
--- a/crux/validate/validate-fk-swap-double-drop.test.ts
+++ b/crux/validate/validate-fk-swap-double-drop.test.ts
@@ -364,6 +364,57 @@ describe('findSwapTargets', () => {
     `);
     expect(targets).toEqual([]);
   });
+
+  it('detects an aliasless UPDATE referencing the table by name in WHERE', () => {
+    // PostgreSQL allows `UPDATE "x" SET ... WHERE x.col = r.id` without an
+    // alias. A migration written this way without the validator's protection
+    // is exactly the failure mode that motivated this check.
+    const targets = findSwapTargets(`
+      UPDATE "publications"
+      SET resource_id = r.stable_id
+      FROM "resources" r
+      WHERE publications.resource_id = r.id;
+    `);
+    expect(targets).toEqual([
+      expect.objectContaining({
+        table: 'publications',
+        column: 'resource_id',
+      }),
+    ]);
+  });
+
+  it('detects an aliasless UPDATE with implicit column reference in WHERE', () => {
+    // `WHERE resource_id = r.id` (no prefix) is also valid in aliasless
+    // form and resolves to the UPDATE target table's column.
+    const targets = findSwapTargets(`
+      UPDATE "publications"
+      SET resource_id = r.stable_id
+      FROM "resources" r
+      WHERE resource_id = r.id;
+    `);
+    expect(targets).toEqual([
+      expect.objectContaining({
+        table: 'publications',
+        column: 'resource_id',
+      }),
+    ]);
+  });
+
+  it('does not span ADD CONSTRAINT detection across statement boundaries', () => {
+    // Two unrelated statements: an ALTER TABLE on table A (no FK swap),
+    // followed by a separate ADD CONSTRAINT on table B that adds the FK
+    // to resources(stable_id). Without the [^;]*? bound, ADD_CONSTRAINT_RE
+    // would attribute the FK to table A. Verify it captures table B.
+    const targets = findSwapTargets(`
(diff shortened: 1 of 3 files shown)


OAGr:
@coderabbitai review

PR #4515, opened 2026-04-20
fix(QUA-618): Microsoft CEO — Süleyman → Nadella

OAGr:
## Summary

- Correct the `ceo` fact on `packages/factbase/data/fb-entities/microsoft.yaml` from **Mustafa Süleyman** to **Satya Nadella**.
- Satya Nadella has been Microsoft's CEO since 2014; Mustafa Süleyman leads the Microsoft AI division (since March 2024), not Microsoft Corp.
- The previous value's source pointed at Microsoft AI's Wikidata item (`Q125891217`) — updated to Nadella's Wikipedia page and added a `notes` clarifying Süleyman's role so the same mis-copy doesn't recur.
- Added `asOf: 2026-04` to match the pattern used by neighboring `ceo` facts (amd, asml, tsmc).

## Scope

Narrow, user-directed fix. A broader refactor is observed but deferred — the same `microsoft.yaml` entity's `parent-organization`, `wikipedia-url`, and Wikidata source all point at Microsoft AI, suggesting the entity-identity itself is muddled. See the 2026-04-20 comment on QUA-618 for the proposed full split. Out of scope for this PR.

## Test plan

- [x] `pnpm crux w validate gate --scope=content --fix` — all 5 content checks pass
- [x] Factual verification — Nadella confirmed as Microsoft CEO since 2014; Süleyman heads Microsoft AI division since March 2024
- [x] Pattern check — value style matches other `ceo` facts in the codebase (plain string, no `!ref`)

Fixes QUA-618

### DIFF
--- a/packages/factbase/data/fb-entities/microsoft.yaml
+++ b/packages/factbase/data/fb-entities/microsoft.yaml
@@ -11,9 +11,10 @@ facts:
     notes: Migrated from old entity system
   - id: f_QYAnAj3qg6
     property: ceo
-    value: !ref sid_DBq1ddu8Mg
-    source: https://www.wikidata.org/wiki/Q125891217
-    notes: From Wikidata Q125891217
+    value: !ref sid_1L16qpyk5A
+    asOf: 2026-04
+    source: https://en.wikipedia.org/wiki/Satya_Nadella
+    notes: Satya Nadella has been Microsoft's CEO since 2014. Previous main value was sid_DBq1ddu8Mg (Mustafa Süleyman), who leads the Microsoft AI division, not Microsoft Corp — corrected per QUA-618.
   - id: f_cLGLs94fuy
     property: parent-organization
     value: Microsoft


packages/factbase/data/fb-entities/microsoft.yaml
```diff
@@ -11,9 +11,10 @@ facts:
     notes: Migrated from old entity system
   - id: f_QYAnAj3qg6
     property: ceo
-    value: Mustafa Süleyman
-    source: https://www.wikidata.org/wiki/Q125891217
-    notes: From Wikidata Q125891217
+    value: Satya Nadella
+    asOf: 2026-04
+    source: https://en.wikipedia.org/wiki/Satya_Nadella
+    notes: Satya Nadella has been Microsoft's CEO since 2014. Mustafa Süleyman leads the Microsoft AI division, not Microsoft Corp.
```
OAGr: This has no need to mention Mustafa! Don't contaminate the raw data with cleanup/fix details.

PR #4527, opened 2026-04-20
feat(QUA-632): /api/enrichment/propose defensive enrichment gate

OAGr:
## Summary

Phase 1 of the [QUA-637](https://linear.app/quantifieduncertainty/issue/QUA-637) coverage-expansion umbrella. Ships the server-side tier-aware gate that enforces "no row without (evidence + verdict)" at a single chokepoint, plus the backing ledger tables.

The gate design is driven by [PR #4518](https://github.com/quantified-uncertainty/longterm-wiki/pull/4518) (QUA-635 verdict-LLM calibration): Haiku 4.5 is 0 % false-confirm at a **strict `verdict === "confirmed"` rule** with ~92 % recall. This PR encodes that rule at the server boundary so future callers can't accidentally weaken it.

## What lands

- **Migrations 0202 / 0203** (note: 0201 was taken by QUA-576; the QUA-632 spec's slot numbers shift by one):
  - `enrichment_targets` — estimated denominators per `(entity_id, record_type)` with a `basis` field so the estimate stays falsifiable. Used by coverage bursts to track progress.
  - `enrichment_runs` — per-burst ledger: proposes_total/accepted/rejected, tier breakdown, verdict distribution, cost. `id` is caller-supplied so a burst can resume across sessions. Has a `CHECK (char_length(id) <= 128)` for defense-in-depth.
- **`POST /api/enrichment/propose`** (`apps/wiki-server/src/routes/enrichment/enrichment.ts`):
  - **T1** — `sourceUrl` must match the T1 authority allowlist for the `(recordType, fieldName)` pair. On match, server produces `sourcing = {verdict: "confirmed", confidence: 1.0, checkedBy: "t1-<source-id>", ...}` — no verdict-LLM call.
  - **T2** — caller supplies `{verdict, quotedText, confidence, checkerModel}`; server enforces `verdict === "confirmed"`, minimum quote length (40 chars), and — when `sourceContent` is supplied — whitespace-tolerant verbatim substring check.
  - **T3** — T2 rules + reject homepage-like URLs (`/`, `/about`, `/contact`, etc.) to prevent the "homepage as source" failure mode from QUA-426 / QUA-648.
  - **Atomic (row + evidence + verdict)** is delegated to the existing grants sync sub-app via an in-process Hono `.fetch()` call. The sync handler's `writeInlineVerdicts` is already transactional; `/propose` does not duplicate the pipeline. Query-string escape hatches (`requireSourcing`, `forceSkipSourcing`, `skipEntityValidation`) are deliberately NOT plumbed through — the whole point of the gate is to enforce these uniformly.
  - Phase 1 supports `grants` as the initial record type; other record types are a one-entry extension to the in-file `RECORD_TYPE_ROUTES` registry once their sync handlers expose a stable route.
- **T1 allowlist** (`t1-allowlist.ts`) seeded with 7 sources: EA Funds, Coefficient Giving, OpenAlex, Wikidata, SEC EDGAR, Hugging Face Leaderboard, GitHub API. Each entry names the record types it's authoritative for.
- **Homepage detector** (`homepage-detector.ts`) — small pure module matching `/`, `/index`, `/home`, `/about`, `/contact` variants; lower-cased path comparison.
- **42 tests** (`apps/wiki-server/src/__tests__/enrichment-propose.test.ts`): T1 allowlist unit tests, homepage-detector table-driven cases, `computeSourcingForTier` pure-function cases (T1/T2/T3 accept + reject variants including `verdict=partial`, short quote, non-verbatim quote, whitespace tolerance), endpoint integration tests with the inner grants route mocked (capturing the forwarded row + sourcing), and ledger UPSERT SQL-capture tests.

## Deferred to follow-up

- **`crux tb tablebase loop` refactor to call `/propose`** (the last bullet of the QUA-632 deliverable). The refactor is an agent-tool change in `crux/tablebase/` that is cleaner to ship independently once this gate is merged and deployed. Filing as a follow-up Linear ticket so the endpoint's first caller lands with full context. No code ships today that exercises `/propose` against prod — the endpoint is available for the follow-up to wire up.
- **Per-field verdicts**: `writeInlineVerdicts` hardcodes `field_name = NULL`. Phase 1 therefore rejects non-null `fieldName` on `/propose` rather than silently writing a row-level verdict. Phase 2 will extend `InlineSourcing` + the sync pipeline to carry `field_name` through.

## Test plan

- [x] `pnpm crux w validate gate --fix` — 53/53 blocking checks pass (3 pre-existing advisory)
- [x] `pnpm build` — clean
- [x] `pnpm --filter "./apps/wiki-server" test` — 1337 passed / 55 skipped (was 1332/55 before; +5 from new tier-gate branches)
- [x] `pnpm --filter "./apps/wiki-server" test -- enrichment-propose` — 42/42
- [x] `npx tsc --noEmit -p apps/wiki-server/tsconfig.json` — clean

## Observed this session

- **Migration slot shift**: QUA-632 spec called for 0201/0202; slot 0201 was already occupied by the merged QUA-576 `resources_stableid_notnull` migration. Using **0202 / 0203** instead. No cross-referencing tickets to update — QUA-632 itself still points at the right thing.
- **PR #4520 (QUA-634/639) references QUA-632** in its PR body as context. No real claim collision; `crux sys agent-checklist init` flagged it as a dedup hit, verified as a false positive (PR is for QUA-634 / QUA-639), forced past with explicit reason.
- **Verdict-LLM stays caller-side (crux), not server-side**. Wiki-server has no LLM infrastructure today; adding it for Phase 1 would be a separate, larger change. The strict-confirmed gate + verbatim quote check + homepage rejection give the same integrity guarantee whether the LLM runs on server or client. When `sourceContent` is supplied, the verbatim check actually verifies the caller's quote against the source — strong defense against a dishonest caller even without server-side LLM.

## Deploy Checklist

- [ ] Post-deploy: verify migration 0202 applied (`\d enrichment_targets`)
- [ ] Post-deploy: verify migration 0203 applied (`\d enrichment_runs`)
- [ ] Post-deploy: `GET /api/enrichment/propose` reachable (expect 405 / 404 because it's POST-only — just confirm the route exists under auth)
- [ ] Post-deploy: sanity-check /internal/data-quality or /health for schema-drift after Drizzle runs both migrations

Fixes QUA-632

### DIFF
--- a/apps/wiki-server/drizzle/0202_qua_632_enrichment_targets.sql
+++ b/apps/wiki-server/drizzle/0202_qua_632_enrichment_targets.sql
@@ -1,0 +1,29 @@
+-- QUA-632 Phase 1: enrichment_targets table.
+--
+-- Stores estimated denominators — per (entity_id, record_type) — so coverage
+-- bursts can track progress toward a percentage goal (e.g. "70% of Anthropic
+-- personnel rows").  `basis` explains how `estimated_total` was derived
+-- (headcount heuristic, LinkedIn employee count, press release, etc.) so the
+-- estimate stays falsifiable.
+--
+-- Companion to 0203_qua_632_enrichment_runs.sql and the /api/enrichment/propose
+-- endpoint.
+
+CREATE TABLE IF NOT EXISTS "enrichment_targets" (
+  "id" bigserial PRIMARY KEY,
+  "entity_id" text NOT NULL,
+  "record_type" text NOT NULL,
+  "estimated_total" integer NOT NULL,
+  "target_pct" real NOT NULL DEFAULT 0.70,
+  "estimated_at" timestamptz NOT NULL DEFAULT NOW(),
+  "basis" text,
+  "confidence" text,
+  "created_at" timestamptz NOT NULL DEFAULT NOW(),
+  "updated_at" timestamptz NOT NULL DEFAULT NOW()
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS "uq_enrichment_targets_entity_record"
+  ON "enrichment_targets" ("entity_id", "record_type");
+
+CREATE INDEX IF NOT EXISTS "idx_enrichment_targets_record_type"
+  ON "enrichment_targets" ("record_type");
--- a/apps/wiki-server/drizzle/0203_qua_632_enrichment_runs.sql
+++ b/apps/wiki-server/drizzle/0203_qua_632_enrichment_runs.sql
@@ -1,0 +1,42 @@
+-- QUA-632 Phase 1: enrichment_runs ledger.
+--
+-- One row per defensive-enrichment burst.  /api/enrichment/propose updates
+-- this ledger as proposes arrive, so we can see tier breakdown, cost, and
+-- verdict distribution per run without joining evidence rows.
+--
+-- `id` is caller-supplied (e.g. a short slug like "anthropic-personnel-t1-2026-04-20")
+-- so the same run can be resumed across sessions.
+
+CREATE TABLE IF NOT EXISTS "enrichment_runs" (
+  "id" text PRIMARY KEY CHECK (char_length("id") > 0 AND char_length("id") <= 128),
+  "label" text,
+  "record_type" text,
+  "entity_id" text,
+  "tier" text,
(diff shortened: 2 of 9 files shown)


OAGr:
@coderabbitai review

PR #4606, opened 2026-04-25
QUA-707: registry-driven frontier-safety-framework ingester

OAGr:
## Summary

Adds the orchestration layer on top of the QUA-691 foundation so the 12 published frontier-safety frameworks (~20 versions in `data/frameworks/frameworks.yaml`) can be ingested end-to-end from the CLI.

- **`crux/frameworks/registry.ts`** — loads + validates the YAML manifest, derives deterministic 10-char framework ids from registry keys, sorts versions by `published_date` and assigns monotonic `versionSortOrder`.
- **`crux/frameworks/sync-payloads.ts`** — extracted `toThresholdSyncItems` / `toDiffSyncPayloads` from `commands/frameworks.ts` so the CLI and the orchestrator can share them without a circular import.
- **`crux/frameworks/orchestrator.ts`** — the end-to-end pipeline: fetch → versionId by natural-key → skip-if-already-synced → extract thresholds → sync version + thresholds → diff consecutive pairs → sync diff + items. Dependency-injected so unit tests run without a network. `--force-reextract` overrides the skip-if-existing check.
- **New CLI subcommands** `list` / `seed` / `ingest` on `crux tb frameworks`. Both `tb frameworks list` (two-word) and `tb frameworks-list` (hyphenated alias) work. The new commands use the canonical `(args, options)` two-arg dispatcher signature so `--apply`, `--only=<key>`, `--max=N`, `--skip-extract`, `--skip-diffs`, `--force-reextract` arrive correctly through `options`.
- **40 new tests** (registry: 22, orchestrator: 18) covering parsing, validation, id determinism, dry-run, skip-if-existing, force-reextract, frameworkKey filter, maxVersions, fetch/extract failures, broken diff chains, summary shape.

Out of scope (intentional, per `.claude/rules/ticket-sizing.md`): actually running the ingester against prod. That's a batch-shaped follow-up — every threshold + diff still needs human review (`humanReviewed=true`, `reviewVerdict` set) before being surfaced publicly.

## Idempotency contract

- Framework ids = `generateId("framework:" + registryKey)` — re-runs upsert.
- Version ids = `generateId("framework-version:" + frameworkId + ":" + label + ":" + contentHash)` — same source content always produces the same id, so re-fetching is safe.
- Threshold ids = `generateId("framework-threshold:" + versionId + ":" + domain + ":" + tierSortOrder)` — re-extraction with the same inputs upserts in place.
- Before extracting, the orchestrator GETs `/api/safety-framework-versions/:id` to skip already-synced versions (saves the LLM bill on incremental re-runs).

## Smoke tests run

- `pnpm crux tb frameworks list` → prints all 12 frameworks / 20 versions correctly.
- `pnpm crux tb frameworks seed` (dry-run) → shows 12 framework rows ready to sync.
- `pnpm crux tb frameworks ingest --only=anthropic-rsp --max=1 --skip-extract --skip-diffs` → fetches Anthropic RSP, computes a deterministic versionId, exits clean.
- `pnpm crux w validate gate` → ✅ 55/55 blocking + 4 pre-existing advisory.
- `npx vitest run --config crux/vitest.config.ts crux/frameworks/ crux/commands/frameworks.test.ts` → 94 tests pass.

## Test plan

- [x] Existing 54 framework tests still pass
- [x] 40 new tests pass (registry parsing/validation/id determinism, orchestrator happy path / dry-run / skip-if-existing / force-reextract / filter / failures / no-diff-on-skipped-side / summary shape)
- [x] CLI smoke tests (list / seed dry-run / ingest --only filter)
- [x] Full validation gate green

Fixes QUA-707

## Summary by CodeRabbit

* **New Features**
  * Added `frameworks-list`, `frameworks-seed`, and `frameworks-ingest` CLI commands for streamlined framework management
  * Implemented registry-driven batch ingestion workflow with support for framework fetching, extraction, and synchronization
  * Added `--json` output formatting and `--apply` flag for safe preview of changes

* **Bug Fixes**
  * Enhanced error handling with proper exit codes for version and diff validation failures

### DIFF
--- a/crux/commands/frameworks.ts
+++ b/crux/commands/frameworks.ts
@@ -435,9 +435,21 @@ async function ingestCommand(
 ): Promise<CommandResult> {
   const log = createLogger(Boolean(options.ci));
 
+  let maxVersions: number | undefined;
+  if (options.max !== undefined) {
+    const n = parseInt(String(options.max), 10);
+    if (!Number.isFinite(n) || n < 1) {
+      return {
+        exitCode: 1,
+        output: `--max must be a positive integer (got "${options.max}")`,
+      };
+    }
+    maxVersions = n;
+  }
+
   const ingestOpts: IngestOptions = {
     frameworkKey: options.only,
-    maxVersions: options.max ? Math.max(1, parseInt(options.max, 10)) : undefined,
+    maxVersions,
     // Default to dry-run unless --apply is given. --dry-run is also accepted
     // as an explicit hint for symmetry with other tooling, but `apply` is
     // the primary opt-in.
--- a/crux/commands/frameworks.ts
+++ b/crux/commands/frameworks.ts
@@ -1,7 +1,8 @@
 /**
- * Frontier-safety-framework subcommands — `crux tb frameworks ...` (QUA-691).
+ * Frontier-safety-framework subcommands — `crux tb frameworks ...`
+ * (QUA-691 foundation, QUA-707 registry-driven ingest).
  *
- * Subcommands:
+ * Per-URL subcommands (manual single-version ingest):
  *   extract <url>  --framework=<sid> --version=<label>
  *                  [--apply]           POST thresholds to wiki-server /sync
  *                  [--model=sonnet]    LLM model override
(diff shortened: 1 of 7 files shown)


OAGr:
🤖 PR Patrol: addressed all 6 CodeRabbit findings in 41a365ef7.

**Major (1)**
- `orchestrator.ts` — dropped `.catch(() => null)` from the `fetchExistingVersion` lookup. The dependency contract documents that it throws loudly on server/auth/unavailable so the caller doesn't re-extract under a false negative; the swallow defeated that and let a 500 / network blip / auth failure trigger duplicate LLM-paid extraction.

**Minor (5)**
- `frameworks.ts` — validate `--max` as a positive integer (was `Math.max(1, parseInt('abc', 10))` → NaN → falsy in orchestrator → silently disabled the cap).
- `orchestrator.ts` — tightened 404 detection in `defaultIngestDeps.fetchExistingVersion` to also require message starts with "404", matching the `adaptIdLookup` pattern from `setup-org.ts`. Prior shape would have absorbed 400/422/429 if classifyStatus ever changed.
- `registry.ts` — keyed `loadRegistry`'s cache by resolved file path. Prior shared singleton silently returned the first-loaded registry on every subsequent call regardless of the `filePath` arg.
- `sync-payloads.ts` — mixed `tierLabel` into the threshold ID seed. The schema has no unique constraint on `(versionId, riskDomainCanonical, tierSortOrder)`, so two distinct triggers in the same domain/tier got the same id and the second silently overwrote the first under upsert.
- `sync-payloads.ts` — replaced positional `idx` in diff-item ID seed with a stable `tierLabel` + `classifierTag` discriminator. Prior shape meant any `structuralDiff` sort change churned the entire row set — orphaning N old rows + inserting N new ones with identical content.

All 86 frameworks tests pass.

(PR's only remaining CI red is Protected Paths — needs human `gate:rules-ok` label since this touches `crux/commands/` + `crux/frameworks/`.)

PR #4613, opened 2026-04-25
QUA-723: flagship-curate reset writes needsRecheck=true so cron picks orphans up

OAGr:
## Summary

`crux flagship-curate` step 4 reset stale verdicts to `unchecked` but wrote them with `needs_recheck=false` and an empty `next_check_due`. The weekly `sourcing-recheck` cron's `getDueForRecheck()` only matches rows where `needs_recheck=true OR next_check_due <= NOW()`, so any reset whose in-process verify (step 5) failed or was cut off by budget became a permanent orphan.

As of 2026-04-24, Anthropic alone had 18 such rows aged 4–11 days; system-wide impact is likely in the 100s.

## Changes

- **`crux/commands/flagship-curate.ts`** — reset write now passes `needsRecheck: true` and `nextCheckDue: new Date().toISOString()`. Either field is sufficient for the cron to pick the row up; setting both is unambiguous and predicate-change-proof. Comment marks the canonical `reasoning` literal as the anchor for the migration.
- **`apps/wiki-server/drizzle/0213_qua_723_flagship_curate_orphan_backfill.sql`** — one-shot UPDATE that re-arms historical orphans. Matches `reasoning` by exact equality (not `ILIKE '%flagship-curate%'`) to avoid colliding with LLM-generated reasoning text from other source-check writers. Wraps in a `DO` block with `RAISE NOTICE` of the affected row count for ops visibility. Idempotent — re-running is a no-op because the WHERE excludes rows whose `next_check_due` has been backfilled.
- **`crux/commands/flagship-curate.test.ts`** — new test asserts the reset payload includes `needsRecheck=true` and a recent `nextCheckDue` ISO string.

## Test plan

- [x] `pnpm vitest run crux/commands/flagship-curate.test.ts` — 59/59 pass (1 new)
- [x] `pnpm build` — exits 0
- [x] `pnpm test` — 7906 pass / 26 skipped, no new failures
- [x] `pnpm crux w validate gate --fix` — 55/55 pass
- [x] `npx tsc --noEmit` on apps/web, apps/wiki-server, and crux flagship-curate files — clean
- [x] Adaptive `/agent-review-pr` (4 files / 76 lines): hostile diff review found 6 findings; 3 fixed (ILIKE→exact, RAISE NOTICE, coupling comment), 2 documented as conservative (issue acceptance pinned WHERE to `IS NULL`), 1 dismissed
- [ ] After merge: monitor next `sourcing-recheck` cron run; expect Anthropic's 18 reset verdicts to flip away from `unchecked`

Fixes QUA-723

## Deploy Checklist

- [ ] `migration` Verify migration 0213 applied on server start — `psql "$DATABASE_URL" -c "SELECT * FROM drizzle.__drizzle_migrations ORDER BY created_at DESC LIMIT 5;"`
- [ ] `manual` Confirm the migration re-armed historical orphans (expect a non-zero `RAISE NOTICE` row count in the deploy logs, then `psql "$DATABASE_URL" -c "SELECT count(*) FROM source_check_verdicts WHERE reasoning='Reset by flagship-curate before re-verification' AND verdict='unchecked' AND next_check_due IS NULL;"` should return 0).
- [ ] `manual` After the next weekly `sourcing-recheck` cron run, verify Anthropic's 18 known orphans have flipped away from `unchecked` — `pnpm crux fb show anthropic` and inspect the verdict mix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where certain source check verdicts were left in an incomplete state; they are now properly re-armed for rechecking.

### DIFF
--- a/apps/wiki-server/drizzle/0213_qua_723_flagship_curate_orphan_backfill.sql
+++ b/apps/wiki-server/drizzle/0213_qua_723_flagship_curate_orphan_backfill.sql
@@ -1,0 +1,27 @@
+-- QUA-723: Backfill flagship-curate reset orphans.
+--
+-- `crux flagship-curate` previously wrote `unchecked` reset verdicts with
+-- `needs_recheck=false` and `next_check_due=NULL`, so the weekly
+-- `sourcing-recheck` cron skipped them permanently. The code path is fixed
+-- in this PR (set both fields on every reset write), but existing orphan
+-- rows need a one-shot rescue so the next cron run picks them up.
+--
+-- Match criteria:
+--   - reasoning ILIKE '%flagship-curate%'  (matches the canonical
+--     "Reset by flagship-curate before re-verification" message)
+--   - verdict = 'unchecked'
+--   - next_check_due IS NULL                (only the orphans, not rows
+--                                            that already have a recheck
+--                                            scheduled)
+--
+-- Idempotent: safe to re-run; the WHERE clause excludes any row whose
+-- next_check_due has already been backfilled.
+
+UPDATE "source_check_verdicts"
+SET
+  "needs_recheck" = true,
+  "next_check_due" = NOW(),
+  "updated_at" = NOW()
+WHERE "reasoning" ILIKE '%flagship-curate%'
+  AND "verdict" = 'unchecked'
+  AND "next_check_due" IS NULL;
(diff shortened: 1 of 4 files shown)


OAGr:
Thanks @coderabbitai. Both nitpicks considered, both declined:

1. **Extract `'Reset by flagship-curate before re-verification'` to a constant** — debatable trade-off. Coupling the migration's SQL literal to a TS file would create a worse coupling than the source-side anchor comment already provides (a future agent renaming the constant in TS would still need to write a follow-up migration). The inline comment + this PR's exact-equality migration WHERE are sufficient.
2. **Stagger `next_check_due` with a random interval** — functionally inert given the actual cron query shape. `getDueForRecheck` uses `OR(next_check_due <= NOW(), needs_recheck = true)` and orders by `priority DESC, lastComputedAt DESC` — `next_check_due` is in neither the predicate's discriminating side (because `needs_recheck=true` is also set on every backfilled row) nor the sort. Load shedding here is governed by the cron's `limit` parameter, not by timestamp jitter.

PR #4621, opened 2026-04-25
QUA-735: fix tb system-cards extract URL parsing (and frameworks)

OAGr:
## Summary

Urgent fix for the QUA-702 (PR #4600) deploy task that always exited with a usage error before extraction could run:

```
pnpm crux tb system-cards extract <url> --ai-model=<sid> --apply
# → Usage: crux tb system-cards extract <url> --ai-model=<entityId> [--apply] [--lab=<key>]
```

### Root cause

`extractCommand` was declared as `(opts: SystemCardOptions)` while the dispatcher (`crux.mjs:512`) passes `(args, options)`. So `opts` was actually the positional-arg array — `opts.url`, `opts.aiModel`, and `opts.apply` were always `undefined`, and the fallback `(opts as { args?: unknown[] }).args` was reading `.args` off an array (also undefined). All three forms tried in QUA-735 (positional URL, `--url=`, flag-first) failed identically before any extraction could begin.

The same anti-pattern existed in `frameworks.ts` (extract / diff / fetch). It happened to work for `tb frameworks <verb>` because `defaultCommand` shimmed `args` onto the options object — but the flat-form aliases `tb frameworks-extract <url>` etc. bypassed the shim and were broken too.

### Fix

- `system-cards.ts::extractCommand` — convert to standard `(args: string[], opts)` signature; URL is parsed from `args.find(a => !a.startsWith('--'))`.
- `system-cards.ts::defaultCommand` — new dispatcher that strips the `extract` verb so both `tb system-cards extract <url>` and `tb system-cards-extract <url>` work identically.
- `frameworks.ts` — same conversion for `extract`, `diff`, and `fetch`. `defaultCommand` no longer needs the `{ ...options, args }` shim; `REGISTRY_DRIVEN` and `LEGACY` tables merge into a single `SUBCOMMANDS` map.

### Verification

Smoke-tested against prod (no extraction actually charged — failures intentional past the validation guards):

```
$ pnpm crux tb system-cards extract https://example.com/nonexistent --ai-model=sid_TEST
… Error: source-card fetch failed: url=… status=dead httpStatus=404
```

Validation guards now reach the extraction layer instead of bailing before it. Same behavior for `--url=` form, flat `system-cards-extract` alias, and `tb frameworks-extract <url>` (now reaches `--framework` validation).

## Test plan

- [x] 9 new arg-parsing tests in `system-cards.test.ts` — positional URL, `--url` flag, missing-arg usage error, flag-before-positional ordering, verb stripping, bare-URL aliases. All 13 tests in the file pass.
- [x] Existing `frameworks.test.ts` (8 tests) still passes
- [x] `pnpm crux w validate gate --fix` — 61/61 checks pass (3 advisory warnings unchanged)
- [x] Manual smoke test of the failing QUA-702 deploy-task command — passes argument parsing and reaches extractor
- [x] Manual smoke test of `tb frameworks extract` and `tb frameworks-extract` flat alias

## Deploy notes

CLI-only change. No migrations, no schema, no runtime impact on the wiki-server. The QUA-702 deploy task ("After deploy: extract a few system cards via `pnpm crux tb system-cards extract …`") can now actually be executed.

Fixes QUA-735.

### DIFF
--- a/crux/commands/frameworks.ts
+++ b/crux/commands/frameworks.ts
@@ -90,11 +90,10 @@ function resolveLlmModel(name?: string): string {
 // Subcommands
 // ---------------------------------------------------------------------------
 
-async function extractCommand(opts: FrameworkOptions): Promise<CommandResult> {
+async function extractCommand(args: string[], opts: FrameworkOptions): Promise<CommandResult> {
   const log = createLogger(Boolean(opts.ci));
 
-  const args = (opts as { args?: unknown[] }).args;
-  const positional = Array.isArray(args) && typeof args[0] === 'string' ? (args[0] as string) : undefined;
+  const positional = args.find((a) => typeof a === 'string' && !a.startsWith('--'));
   const url = opts.url ?? positional;
   if (!url) {
     log.error(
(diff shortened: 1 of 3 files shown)


OAGr:
🤖 PR Patrol: the failing `test` job is a **flake unrelated to this PR**. The failure was `lib/career-import > produces deterministic IDs across runs` — a 5s test timeout. This PR only touches `crux/commands/{frameworks,system-cards*}.ts`, nothing in `lib/career-import/`. The same test passes on main (run 24938281732). Filed [QUA-738](https://linear.app/quantifieduncertainty/issue/QUA-738) for the underlying flake (the test runs `extractAllCareers()` twice on the full corpus inside the default 5s timeout — too tight when CI is loaded).

Re-running the `test` job should pass. Protected Paths is the other failure — needs human `gate:rules-ok` label.

PR #4650, opened 2026-04-28
fix(QUA-684): render date-shaped integers as dates, not magnitudes

OAGr:
## Summary

- New `formatDateShapedInteger` in `apps/web/src/lib/format-compact.ts` detects 8/12/14-digit YYYYMMDD[hhmm[ss]] integers (year 1900-2099, calendar-valid month/day/time, Date round-trip — rejects Feb 30, Feb 29 in non-leap years, etc.) and renders them as human dates.
- Wired into three call sites:
  - `formatFactValueString` (the fact `value` column on the `/organizations/*` Database tab) — `"20240601000000"` now renders as `Jun 1, 2024 00:00:00 UTC` instead of `$20.2T`.
  - The generic `typeof value === "number"` path in the entity-profile cell renderer.
  - `sanitizeRawLargeNumbers` (description columns) — embedded date-shaped runs route to date format instead of compact magnitude.

## Why

QUA-673 + QUA-675 healed the original symptom on prod (URL-embedded timestamps in `notes` columns). This PR closes the latent class flagged in the QUA-684 issue: any BIGINT timestamp column that ever surfaces a value like `20240601000000` on the Database tab will render as a date, not a misleading magnitude.

9/10/11/13-digit Unix epochs (seconds and milliseconds) are intentionally NOT detected — they overlap the plausible-magnitude range for headcount and revenue facts. The QUA-684 issue explicitly flagged these as too ambiguous.

## Test plan

- [x] 89 unit tests in `format-compact.test.ts` and `format-cell-value.test.ts` (62 + 26 + 1 new century-leap test): happy paths for 8/12/14 digit shapes, year/month/day/hour/minute/second bounds, Gregorian century-year leap rule (1900 not leap, 2000 leap), Feb 29 in non-leap years, NaN/Infinity/non-integer/negative rejection, render-audit `\d{10,}` regex compliance.
- [x] Adversarial fuzz against `formatDateShapedInteger` with 20 boundary inputs (2^53, scientific notation results, hour 24, minute 60, century leap edge cases) — all behave correctly.
- [x] Full apps/web suite (`pnpm vitest run`): 1745 tests pass.
- [x] `pnpm build`: 0 errors.
- [x] Full gate (`pnpm crux w validate gate --fix`): 62 blocking checks pass.
- [x] Render audit against prod (`PLAYWRIGHT_BASE_URL=https://www.longtermwiki.com npx playwright test e2e/render-audit.spec.ts`): 30/30 pass.

## Reviewer notes

- The Date round-trip check is novel to this codebase — `formatIntroducedDate` and friends only do range-validity, not Feb-29-in-non-leap-year detection. Worth keeping local for now; if a second caller appears, extract `isValidUtcDate(year, month, day)` as a follow-up.
- Range gate `n < 19000101 || n > 20991231235959` skips the `String(n)` allocation for the vast majority of cell-render calls (small counts, large magnitudes outside the date envelope) — matters because the function runs per cell per row.

Closes QUA-684

### DIFF
--- a/apps/web/src/app/internal/entity-profile/__tests__/format-cell-value.test.ts
+++ b/apps/web/src/app/internal/entity-profile/__tests__/format-cell-value.test.ts
@@ -95,6 +95,51 @@ describe("formatFactValueString (QUA-673)", () => {
   });
 });
 
+describe("formatFactValueString — date-shaped integers (QUA-684)", () => {
+  it("renders 14-digit YYYYMMDDhhmmss as a date, not a magnitude", () => {
+    // Without the date-shape check this would render as "$20.2T" — the
+    // openai render-audit-flagged value 20240601000000 (a Wayback Machine
+    // timestamp) being mistaken for a financial magnitude.
+    expect(formatFactValueString("20240601000000", "USD")).toBe(
+      "Jun 1, 2024 00:00:00 UTC"
+    );
+    expect(formatFactValueString("20240601000000", null)).toBe(
+      "Jun 1, 2024 00:00:00 UTC"
+    );
+  });
+
+  it("renders 12-digit YYYYMMDDhhmm as a date with hh:mm UTC", () => {
+    expect(formatFactValueString("202406011830", "USD")).toBe(
+      "Jun 1, 2024 18:30 UTC"
+    );
+  });
+
+  it("renders 8-digit YYYYMMDD as a date with no time", () => {
+    expect(formatFactValueString("20240601", null)).toBe("Jun 1, 2024");
+  });
+
+  it("date-shape check fires regardless of the 1000-magnitude floor", () => {
+    // 19500101 = 19.5M — well above the floor anyway, but the test asserts
+    // that the date branch wins even when both branches would otherwise fire.
+    expect(formatFactValueString("19500101", "USD")).toBe("Jan 1, 1950");
+    // 20240229 (Feb 29 of a leap year) — calendar-valid date.
+    expect(formatFactValueString("20240229", null)).toBe("Feb 29, 2024");
+  });
+
+  it("8-digit numbers that aren't valid dates stay raw or compact-formatted", () => {
+    // 19000000 has month=00 → not a date → falls through to magnitude path → "19M".
+    expect(formatFactValueString("19000000", null)).toBe("19M");
+    // 12345678 — not a date (12=month=12 valid? yes 12; day=34 invalid → not a date) → "12M".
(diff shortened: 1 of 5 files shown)


OAGr:
Closing without merge. Investigation found this PR fixes a problem that doesn't exist in current production data, with a heuristic that has real false-positive risk.

## Findings

**The "BIGINT timestamp columns" claim is fictional.** Grepped every `bigint(...)` declaration in `apps/wiki-server/src/schema.ts`. There are zero BIGINT-as-timestamp columns. All timestamp columns use `timestamp(..., { withTimezone: true })`. The storage problem the docstring describes doesn't exist in this codebase.

**No FactBase fact has a date-shaped value.** Scanned all 511 facts in `factbase-data.json` plus every `records.<sid>.<table>[N].fields.value` path. Zero matches against `^(?:19|20)\d{6}(?:\d{4}|\d{6})?$`. The unit test `formatFactValueString("20240601000000", "USD")` is purely hypothetical.

**The actual `20240601000000` value is a Wayback URL inside `notes` fields.** 72 grants (all from coefficient-giving) have notes like `https://web.archive.org/web/20240601000000/https://openphilanthropy.org/grants/...`. The `notes` column takes the long-text path through `linkifyText` — `sanitizeRawLargeNumbers` is **not** called on it.

**QUA-684 is already fixed.** Ran the render-audit against prod (`PLAYWRIGHT_BASE_URL=https://www.longtermwiki.com npx playwright test e2e/render-audit.spec.ts -g openai`) — passes. QUA-675 (PR #4570) shipped URL linkification on Apr 24, putting the timestamp inside `<a href>` where it doesn't appear in `innerText`. Browser-side check on `/organizations/openai` confirms `20240601000000` is not in visible text.

## Why not merge anyway as defense-in-depth

The date-shape detector cannot distinguish `20100201` interpreted as "Feb 1, 2010" from the count "20,100,201". It deliberately skips 9/10/11/13-digit Unix epochs because they're "ambiguous" with magnitudes — but those are the more common real-world timestamp encodings. Adding the heuristic now means a future numeric fact whose value happens to land in the YYYYMMDD envelope silently renders as a date. Brittle for a benefit that pays out in zero current cases.

## Better path if a defense-in-depth layer is wanted

Make `sanitizeRawLargeNumbers` URL-aware: split on `https?://...`, sanitize only non-URL spans, then linkify. URLs are a known structural unit; treat them as such instead of guessing at value shapes. Generalizes to request IDs, hashes, version strings, and anything else that lives inside URLs.

Closing QUA-684 as resolved by QUA-675.

PR #4736, opened 2026-04-30
QUA-927: url-resolves verifier — cheap path for URL-only fact properties

OAGr:
## Summary

Adds a third verifier kind alongside `content-claim` — `url-resolves` — for the cheap, $0 verification path described in QUA-927. Properties whose stored value IS the URL (Wikipedia article, GitHub profile, Google Scholar, social media) no longer require a content-claim LLM check.

Closes QUA-927.

## Design — verifier rides the Resource pipeline

**The verifier does not fetch URLs.** It consumes cached state from the existing Resource pipeline and self-heals via `suggestResourcesApi` for unknown URLs.

> History: an earlier draft of this PR issued its own HEAD/GET against the live URL with parallel SSRF / timeout / redirect / Wikipedia HTML-parse code. That bypassed `content_hash` change detection (QUA-312/329), didn't produce a `resources` row, and duplicated logic the resource-ingest worker already runs. Switched to Resource-routed in `2094d5f16` per review.

Decision tree:

| Cached state | Verdict | Confidence |
|---|---|---|
| `citation_content.httpStatus` 2xx (non-Wikipedia) | `confirmed` | 0.95 |
| `citation_content.httpStatus` 2xx + Wikipedia pageTitle matches entity | `confirmed` | 0.95 |
| `citation_content.httpStatus` 2xx + Wikipedia pageTitle does NOT match | `contradicted` | 0.85 |
| `citation_content.httpStatus` 4xx/5xx | `contradicted` | 0.95 |
| `resources.fetchStatus` is dead/soft_404/not_found/timeout/unreachable | `contradicted` | 0.95 |
| `resources.fetchStatus` is `paywall` | `confirmed` (URL alive, body gated) | 0.85 |
| `resources.fetchStatus` is `ok` (no content yet) | `confirmed` | 0.85 |
| Nothing cached anywhere | `unverifiable` + enqueue `resource-ingest` job | 0.7 |

Self-healing: a brand-new URL returns `unverifiable` on its first pass and enqueues ingest. The next pass (after the worker runs) gets a real verdict from the cached state. Same flow `source-fetcher.ts` already uses for content-claim.

## What changed

| File | Change |
|------|--------|
| `crux/lib/sourcing/url-resolves-verifier.ts` | Resource-pipeline consumer (~290 lines, was ~535) |
| `crux/lib/sourcing/url-resolves-verifier.test.ts` | 46 tests mocking `lookupResourceByUrl`, `getCitationContentByUrl`, `suggestResourcesApi` |
| `crux/lib/sourcing/item-verifier.ts` | Dispatch url-resolves first, before deterministic matchers |
| `crux/lib/sourcing/orchestrator.ts` | Inline url-resolves in the batch path so these items never queue for the Anthropic Batch API |
| `crux/lib/sourcing/item-collectors.ts` | Stop skipping facts whose property has `verifierKind: url-resolves`; propagate `propertyId` and `verifierKind` onto `FactItemData` |
| `crux/lib/sourcing/orchestrator-types.ts` | New optional fields on `FactItemData`: `propertyId`, `verifierKind` |
| `packages/factbase/src/types.ts` | New `Property.verifierKind` field |
| `packages/factbase/data/properties.yaml` | Mark `wikipedia-url`, `social-media`, `github-profile`, `google-scholar` as `verifierKind: url-resolves` (drops `verifiable: false` on Wikipedia and social-media) |

Cost: $0 per check. Latency: one wiki-server lookup per check (~10–50ms) for known URLs.

## Why route through Resources

- **Content-hash change detection** (QUA-312/329) — a Wikipedia URL rewritten to a different topic flips the `content_hash`, triggers re-enrichment, triggers a fresh source-check verdict. Direct HEAD has no temporal awareness.
- **`resources` row exists** — `archive_url`, `last_fetched_at`, `fetch_status` available to admins. Verdict is no longer a parallel universe.
- **Wayback fallback** for dead links — `source-fetcher.ts` already does this; the url-resolves path now inherits it.
- **One HTTP code path** — SSRF guard, timeout, redirect, HTML title extraction all live in the resource-ingest worker. One place to fix bugs.

## Acceptance criteria — from QUA-927

- [x] New `url-resolves` verifier type implemented + tested (46 tests)
- [x] `wikipedia-url`, `github-profile`, `google-scholar`, `social-media` marked `verifierKind: url-resolves`
- [x] These facts are no longer excluded by `collectFactItems`
- [x] Verdicts land in `source_check_verdicts` with verdict=`confirmed` for valid URLs (uses existing `storeResult` path; `checkerModel='url-resolves'`)

The "expect coverage ceiling rises by ~14pp" criterion is verifiable post-merge by running `pnpm crux fb sourcing --type=fact --table=fact --limit=500` against the now-eligible ~388 facts. First run will enqueue ingest jobs for any not-yet-cached URLs; second run reads the verdicts.

## Test plan

- [x] `npx vitest run crux/lib/sourcing/url-resolves-verifier.test.ts` — 46/46 pass
- [x] `npx vitest run crux/lib/sourcing/` — all 630 sourcing tests pass
- [x] `npx tsc --noEmit -p crux/tsconfig.json` — clean for changed files
- [x] `npx tsc --noEmit -p apps/web/tsconfig.json` — clean
- [x] `npx tsc --noEmit -p apps/wiki-server/tsconfig.json` — clean
- [x] `pnpm crux w validate gate --fix` — all 55 checks pass

## Notes for reviewers

- **`website` was deliberately not included.** The ticket explicitly defers it ("optionally `website` after policy review"). A website fact is still self-referential and there's no easy "right page" check; left for a follow-up.
- **Verifier dispatch ordering**: url-resolves runs *before* the Wikidata and OpenAlex deterministic matchers. The matchers wouldn't match a url-resolves fact anyway (different propertyId), but the explicit ordering keeps the cheapest path first.
- **`verifiable: false` semantics preserved**: legacy facts with `verifiable: false` and no `verifierKind` are still skipped — the bypass is gated on the explicit `verifierKind: 'url-resolves'`.
- **First-pass behavior on new URLs**: returns `unverifiable` and enqueues ingest. Second pass gets a real verdict. Operators running a one-shot sourcing pass against a fresh dataset should expect to run twice.

## Commits

- `2cce11fac` — initial implementation (direct HEAD probe path)
- `1d0b7b7ad` — review fixes: SSRF guard, word-boundary match, robust HTML entity decoder, body-leak fix, credential URL block
- `2094d5f16` — **rewrite to consume Resource pipeline** per review feedback (drops 508 lines net)

### DIFF
--- a/crux/lib/sourcing/item-collectors.ts
+++ b/crux/lib/sourcing/item-collectors.ts
@@ -191,8 +191,12 @@ export function collectFactItems(
       if (!fact.source || fact.id.startsWith('inv_')) continue;
 
       const property = graph.getProperty(fact.propertyId);
-      // Skip properties marked as not verifiable (e.g., social media handles, self-referential URLs)
-      if (property?.verifiable === false) continue;
+      // QUA-927: properties marked as `verifierKind: url-resolves` get a cheap
+      // HEAD-only check, bypassing the `verifiable: false` skip even if it was
+      // historically set (e.g., wikipedia-url / social-media). The url-resolves
+      // verifier kicks in below in item-verifier.ts based on the propagated
+      // verifierKind on FactItemData.
+      if (property?.verifiable === false && property?.verifierKind !== 'url-resolves') continue;
       const formattedValue = formatFactValue(fact, property, graph);
       const existing = existingVerdicts.get(fact.id);
 
@@ -213,8 +217,10 @@ export function collectFactItems(
           entity,
           fact,
           propertyName: property?.name ?? fact.propertyId,
+          propertyId: fact.propertyId,
           formattedValue,
           rawValue: extractRawValue(fact),
+          verifierKind: property?.verifierKind,
         },
       });
     }
--- a/crux/lib/sourcing/item-verifier.ts
+++ b/crux/lib/sourcing/item-verifier.ts
@@ -25,6 +25,7 @@ import {
 import { matchRecordAgainstSnapshot } from './deterministic-matcher.ts';
 import { tryWikidataMatch } from './wikidata-matcher.ts';
 import { tryOpenAlexMatch } from './openalex-matcher.ts';
+import { tryUrlResolvesVerify } from './url-resolves-verifier.ts';
 import { searchForEntity } from './item-collectors.ts';
 import { isRelevanceGateEnabled, runRelevanceGate } from './relevance-gate.ts';
 import type {
@@ -286,6 +287,20 @@ export async function verifySingleItem(
   client: ReturnType<typeof createLlmClient>,
   useWebSearch: boolean,
 ): Promise<VerifyResult | VerifyError> {
+  // ── URL-resolves verifier (QUA-927) ──
+  // Cheapest path: facts with `verifierKind: url-resolves` are verified by an
+  // HTTP HEAD request, no LLM, no source-content fetch. Returns null when the
+  // fact isn't tagged for url-resolves verification, so this is safe to run
+  // unconditionally.
+  if (item.data.kind === 'fact' && item.data.verifierKind === 'url-resolves') {
+    try {
+      const urlResolvesResult = await tryUrlResolvesVerify(item);
+      if (urlResolvesResult) return urlResolvesResult;
+    } catch (e: unknown) {
+      console.warn(`[sourcing] url-resolves verifier failed for ${item.id}: ${e instanceof Error ? e.message : String(e)}`);
+    }
+  }
+
   // ── Deterministic matching path (Discussion #3567 Phase 3) ──
   // For any record-type item, try deterministic row-matching against a source
   // snapshot before falling back to LLM sourcing. tryDeterministicMatch
--- a/crux/lib/sourcing/orchestrator-types.ts
+++ b/crux/lib/sourcing/orchestrator-types.ts
@@ -62,9 +62,21 @@ export interface FactItemData {
   entity: FBEntity;
   fact: Fact;
   propertyName: string;
+  /**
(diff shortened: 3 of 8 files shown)


OAGr:
This seems to check the web page directly when verifying. I assumed that instead, we'd ensure that such page is downloaded as a Resource

OAGr:
Good call — the direct HEAD probe was the wrong design. Pushed `2094d5f16` rerouting the verifier through the Resource pipeline.

The verifier no longer fetches URLs. It reads cached state from `lookupResourceByUrl` + `getCitationContentByUrl`, and for unknown URLs enqueues a `resource-ingest` job via `suggestResourcesApi` and returns `unverifiable` for that pass. Same self-healing pattern `source-fetcher.ts` already uses.

Decision tree:
1. `citation_content.httpStatus` 2xx → `confirmed` (Wikipedia: also check `pageTitle` contains entity name)
2. `citation_content.httpStatus` 4xx/5xx → `contradicted`
3. `resources.fetchStatus` dead/soft_404/timeout/etc. → `contradicted`
4. `resources.fetchStatus` paywall → `confirmed` (URL alive, body gated)
5. `resources.fetchStatus` ok → `confirmed`
6. Nothing cached → `unverifiable` + enqueue ingest

Net: -534 lines verifier, -709 lines tests. SSRF guard, timeout/redirect handling, Wikipedia HTML title parsing, HTML-entity decoding — all gone from this file. They already exist in the resource-ingest worker, which is where the URL is actually fetched. We now get content_hash change detection, archive_url, Wayback fallback, and a resources row for free.

46/46 verifier tests pass, 630/630 sourcing tests pass, gate green.

PR #4755, opened 2026-05-01
fix(factbase): backfill-sources batch — reconcile dropped + unknown customIds (QUA-963 HIGH)

OAGr:
## Summary

Fixes the **HIGH finding 2** from QUA-963's hostile-review punch list. PR-B in the QUA-963 sequence (PR-A was #4752, the CRITICAL verify-chain fix; this is the next blocker for safe `--batch --apply` use).

`discoverBatch` silently dropped work in two symmetric cases:

1. **Anthropic returned a customId we didn't submit** (provider drift / response-set corruption): logged a warning + `continue`d without counting.
2. **We submitted a customId Anthropic never returned** (provider drop): never counted at all. `submitted.length=50, responses.size=49` → `summary.scanned=49`, no warning, no error trail. The dropped fact was untraceable.

A `--batch --apply` run could leave dozens of facts unverified with no signal which were affected.

## Fix

- Track `seenCustomIds` during the response loop.
- After the loop, walk `customIdToFact` for unseen customIds and emit a synthetic `FactDiscovery` (`kind: error`) per dropped fact + `summary.engineErrors++` + `summary.missingResponses++`.
- Bump unknown-customId path to `summary.unknownCustomIds++` (separate counter so operators can tell "we lost work" from "they sent garbage").
- Reconciliation log line surfaces totals when either is non-zero.
- Report formatter shows the new fields when > 0 (hidden in real-time mode where they're always 0).

## Tests

3 new tests in addition to the 19 existing:
- All facts dropped (every customId missing from response set) — assert `missingResponses == submitted` and `scanned == submitted` (NOT `responses.size`).
- Unknown customIds returned with full submitted set still valid — assert `unknownCustomIds: 1, missingResponses: 0`.
- Mixed case (1 valid + 1 unknown + 1 dropped) — assert distinct counters track correctly.

Test count: 19 → 22. All pass locally + gate green (62/62 checks).

## Test plan

- [x] Build green
- [x] Gate green (62/62 + 6 triaged + 2 advisory)
- [x] 22/22 unit tests pass
- [x] Type check green on `crux/`
- [x] No-op detection from PR-A (#4752) preserved (regression test still passes)

Closes part of QUA-963 (HIGH finding 2 — finding 1 = CRITICAL = closed via #4752; finding 3 = cost-drift = remains open for PR-C).

Fixes QUA-963

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for batch processing reconciliation scenarios, including missing responses and unexpected API outputs.

* **Bug Fixes**
  * Improved batch operation error handling to properly track and reconcile submitted requests against received responses.
  * Enhanced error reporting with accurate per-fact records and consistent error counting for missing or unexpected responses.

### DIFF
--- a/crux/commands/factbase-backfill-sources.test.ts
+++ b/crux/commands/factbase-backfill-sources.test.ts
@@ -727,6 +727,18 @@ describe('crux fb backfill-sources — --batch path', () => {
     // Post-fix, the wrapper walks customIdToFact at the end of the response
     // loop and emits an engineError + missingResponses++ + a synthetic
     // FactDiscovery so the report can name the dropped fact.
+    //
+    // CodeRabbit fix: the test previously hardcoded `expect(missingResponses).toBe(2)`,
+    // which would fail if the KB had < 2 unsourced born-year facts (e.g.,
+    // after the QUA-933 smoke test wrote a Wikipedia source for Andrej
+    // Karpathy's born-year). Resolve the actual count up-front via
+    // collectUnsourcedFacts and assert against `n`, capping at 2 to keep
+    // the test bounded.
+    const kb = await loadGraphFull();
+    const available = collectUnsourcedFacts(kb, { property: 'born-year', limit: 2 });
+    if (available.length === 0) return; // skip if no born-year facts at all
+    const n = available.length;
+
     mockSubmitBatch.mockResolvedValue({ id: 'batch_drop', processing_status: 'in_progress', request_counts: {} });
     mockPollBatch.mockResolvedValue({ id: 'batch_drop', processing_status: 'ended', request_counts: {} });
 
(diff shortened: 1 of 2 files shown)


OAGr:
Both CodeRabbit findings addressed in commit ed71031eb:

1. **KB-fragility in tests**: replaced hardcoded `expect(missingResponses).toBe(2)` with dynamic count from `collectUnsourcedFacts`. Tests now skip if KB doesn't have enough unsourced born-year facts available, asserting against the actual returned count instead. Resilient to future backfills consuming additional born-year facts.

2. **Cost-charging timing**: moved `summary.costUsd += ...` from per-response success branch to up-front `requests.length * ESTIMATED_BATCH_COST_USD` assignment after `getBatchResults` returns. Now matches submission semantics regardless of drop/error rate. Pinned by new test 'charges batch cost up-front based on submission count, not on successes'.

Test count: 22 → 23. All pass locally.

PR #4767, opened 2026-05-01
QUA-964: fail-loud syncPolicyStakeholders + strict enum gate validator

OAGr:
## Summary

Closes the QUA-941 vulnerability class (silent stakeholder drop) by:

1. **Fail-loud `syncPolicyStakeholders`** — `apps/web/scripts/lib/wiki-server-data.mjs` now distinguishes FK-missing 400s (legitimate skip — some YAML stakeholders reference entities not yet in PG) from any other 400 (data corruption). Non-FK errors print the offending policy ID + response body to stderr and `process.exit(1)`.
2. **Strict gate validator** — `crux/validate/validate-policy-stakeholders-strict.ts` runs the canonical sync-route Zod schema against every `policy.stakeholders[]` block in `data/entities/` at PR-review time, so a YAML PR introducing `position: reform` (or any other route-rejected enum) fails `pnpm crux w validate gate` instead of getting silently dropped at build-data time.
3. **Schema extracted to standalone file** — `apps/wiki-server/src/routes/tablebase/policy-stakeholders-schema.ts` holds `SyncStakeholderItemSchema`, `VALID_POSITIONS`, `VALID_IMPORTANCE`. The route imports from it; the validator imports from it. Single source of truth — any future schema tightening at the route fires the gate validator automatically.

## Acceptance per ticket

- ✅ `pnpm crux w validate gate` fails on a YAML PR that introduces `position: reform` or any other non-allowed enum (verified manually by injecting + running the validator).
- ✅ A sync route returning `{error: "validation_error", ...}` for a non-FK reason now causes `pnpm build` to exit 1 with the offending policy ID in stderr.
- ✅ Zero `position: reform` rows in `data/entities/responses.yaml` (already cleaned by QUA-941 hotfix #4744 — re-verified: 244 stakeholders across 39 policies, all canonical positions).
- ✅ Existing FK-missing 400s still produce `skipped` counts (matched by `body.includes('Entity references not found')`).

## Why this is independent of QUA-943's full v4 cutover

The QUA-941 mechanism is a 10-line `skipped++` bug + a missing strict-schema import at PR time. Neither requires PG-primary writes. v4 (Phases 0–4) is the structural fix; this PR is the tourniquet that closes the immediate vulnerability while v4 proceeds on its own pace.

## Soft dependency on QUA-951 (validationError discriminator)

The fail-loud check uses string-matching on `"Entity references not found"` to detect FK errors. Once QUA-951 lands the `code:` discriminator, the check switches to `body.includes('"code":"fk_missing"')` — TODO comment in place at the relevant line.

## Review fixes (commit f763ff35)

`/agent-review-pr` flagged 2 HIGH + 1 MEDIUM that landed in the review commit:

- **HIGH**: validator's `buildSyncItem` did NOT mirror the helper's `||` semantics — `typeof === "string"` checks coerced non-string `entityId`/`reason`/`source`/`context` to null, so YAML like `entityId: 42` would silently pass the gate while the runtime helper would now correctly fail loud. Fixed: validator passes values through unchanged, Zod catches drift at both sites.
- **HIGH**: only `runCheck()` against current passing YAML was tested. Split into pure `evaluatePolicies(policies)` + thin `runCheck` wrapper. 8 new fixture tests drive the negative path: `position: reform`, missing name, numeric `entityId`, string-typed `context`, missing stableId, multi-issue policies.
- **MEDIUM**: in `syncPolicyStakeholders`, the outer try/catch swallowed `r.text()` throws (interrupted body stream after a non-2xx) and silently incremented `skipped` — exactly the QUA-941 silent-drop pattern. Fixed: split fetch + body-read error handling so only network errors on `fetch()` skip; a 4xx with unreadable body fails loud.

## Test plan

- [x] `npx tsx crux/validate/validate-policy-stakeholders-strict.ts` against current YAML — passes (244 stakeholders, 39 policies)
- [x] Same with one row mutated to `position: reform` — fails with `Invalid enum value` finding and exit 1
- [x] Same with `entityId: 42` injected — fails with `Expected string, received number` (regression test for the helper-vs-validator drift fix)
- [x] `crux/validate/validate-policy-stakeholders-strict.test.ts` — 16 unit tests (8 schema-level, 8 fixture-driven negative paths)
- [x] `pnpm test` (apps/wiki-server) — 1555 pass (mount-registry covers the new schema file)
- [x] `pnpm test` (apps/web) — 1904 pass
- [x] `tsc --noEmit` (apps/web + apps/wiki-server) — clean
- [x] `pnpm crux w validate gate --fix --no-cache` — 70/70 checks pass (2 unrelated advisories: crux baseline tsc, orphan entities)

Closes QUA-964.

### DIFF
--- a/apps/web/scripts/lib/wiki-server-data.mjs
+++ b/apps/web/scripts/lib/wiki-server-data.mjs
@@ -955,36 +955,55 @@ export async function syncPolicyStakeholders(typedEntities) {
       } else {
         // Batch failed — try items individually to skip bad FK references
         for (const item of batch) {
+          // Phase 1: send the request. Network/transport failures are the only
+          // legitimate skip-on-throw; anything else must fail loud.
+          let r;
           try {
-            const r = await fetch(`${serverUrl}/api/policy-stakeholders/sync?${skipSourcing}`, {
+            r = await fetch(`${serverUrl}/api/policy-stakeholders/sync?${skipSourcing}`, {
               method: 'POST',
               headers,
               body: JSON.stringify({ items: [item] }),
               signal: AbortSignal.timeout(10_000),
             });
-            if (r.ok) {
-              synced++;
-            } else {
-              // QUA-941 / QUA-964: distinguish FK-missing (legitimate skip — some
-              // YAML stakeholders reference entities not yet in PG) from Zod schema
-              // rejection (data corruption — must fail loud). FK errors from
-              // validateEntityRefs always contain the literal "Entity references
-              // not found"; any other 400 indicates real schema violation.
-              // TODO(QUA-951): switch to body.includes('"code":"fk_missing"') once
-              // the validationError discriminator lands.
-              const body = await r.text();
-              const isFkMissing = body.includes('Entity references not found');
-              if (!isFkMissing) {
-                process.stderr.write(
-                  `syncPolicyStakeholders: non-FK 400 for policy=${item.policyEntityId} stakeholder=${item.stakeholderEntityId ?? item.stakeholderDisplayName} — YAML data corruption (QUA-941).\nResponse: ${body}\n`,
-                );
-                process.exit(1);
-              }
-              skipped++;
-            }
           } catch {
             skipped++;
+            continue;
           }
+          if (r.ok) {
+            synced++;
+            continue;
+          }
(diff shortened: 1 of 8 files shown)


OAGr:

PR #4768, opened 2026-05-01
QUA-952: emit code: discriminators on canary's 3 routes

OAGr:
## Summary

Phase 0a-ii of QUA-943 (machine-write PG-primary transition). Adds structured `code:` discriminators to the three `validationError` sites the canary `policy.stakeholders` route touches, so Phase 2 dual-write retry-with-feedback (Phase 3) can dispatch on `code` instead of pattern-matching `message`.

- **`sync-factory.ts`** — Zod return path now emits `code: "enum_violation"` (with `field`/`value`/`allowed[]`) when the first issue is `invalid_enum_value`, otherwise `code: "zod"` (with `field`/`value` when `received` is available). policy-stakeholders inherits this through `createSyncHandler` — no per-route changes.
- **`validate-entity-refs.ts`** — FK miss now emits `code: "fk_missing"` with the first missing field's `field`/`value` slots; full multi-field summary stays in `message`. The structured `value` array caps at 50 IDs (`FK_MISSING_VALUE_CAP`) so a 500-item all-invalid batch can't push 10–50KB into the response body.
- **shared helper `zodErrorToValidationBody`** in `routes/shared/utils.ts` centralizes the ZodError → ValidationErrorBody mapping for Phase 5 expansion.

Other 156+ legacy 1-arg `validationError(c, "...")` callsites (the `zv()` query-param path, the natural-key collision path, etc.) are intentionally unchanged per QUA-943 v4 RFC §0a-ii.

**Scope note.** `validate-entity-refs.ts` is shared infrastructure used by ~all sync routes, so the new `code: "fk_missing"` discriminator surfaces on every route that calls `validateEntityRefs`, not just the canary. The change is additive (legacy `message` is preserved verbatim), so existing string-matching consumers are unaffected — but a future Phase 3 retry consumer dispatching on `code: "fk_missing"` will trigger across all sync routes, not just `policy-stakeholders`.

## Test plan

- [x] `npx vitest run validation-error-code-discriminator.test.ts` — 11 new specs (6 integration + 5 direct unit on `zodErrorToValidationBody`) all pass
- [x] `npx vitest run policy-stakeholders.test.ts validate-entity-refs.test.ts sourcing-enforcement.test.ts` — 49 specs across all affected suites pass
- [x] `npx tsc --noEmit -p apps/wiki-server/tsconfig.json` and `apps/web/tsconfig.json` — both clean
- [x] `pnpm crux w validate gate --fix` — 69 checks pass (2 pre-existing advisory warnings)
- [x] `/agent-review-pr` — hostile reviewer subagent ran; 1 HIGH + 5 MEDIUM findings addressed (cap on FK value array, JSON-safety contract docstring, corrected QUA-940 comment, direct unit tests for the helper)
- [x] Adversarial inputs (empty path, number-only path, dotted-key, multi-issue, big `allowed[]`, numeric `received`) hand-tested against `zodErrorToValidationBody`

Fixes QUA-952

### DIFF
--- a/apps/wiki-server/src/__tests__/validation-error-code-discriminator.test.ts
+++ b/apps/wiki-server/src/__tests__/validation-error-code-discriminator.test.ts
@@ -1,0 +1,292 @@
+/**
+ * QUA-952 (Phase 0a-ii) — canary `code:` discriminator response shapes.
+ *
+ * Asserts that the 3 canary routes emit structured `code:` discriminators in
+ * their 400 responses so Phase 2 dual-write retry-with-feedback (Phase 3) can
+ * dispatch on `code` instead of pattern-matching `message`:
+ *
+ *   1. `policy-stakeholders` Zod parse → `code: "zod"` or `code: "enum_violation"`
+ *      (emitted by sync-factory's Zod return path; bullet 3 of the ticket).
+ *   2. `validate-entity-refs` FK miss → `code: "fk_missing"` (bullet 2).
+ *   3. Sync-factory Zod parse fields populate `field`/`value`/`allowed` so a
+ *      retry consumer can re-prompt with the rejected input (bullet 3).
+ *
+ * Other 156+ legacy 1-arg `validationError(c, "...")` callsites are NOT in
+ * scope for QUA-952 — see QUA-943 v4 RFC §0a-ii. Their response shapes are
+ * unchanged and intentionally untested here so this file stays focused on
+ * the canary contract.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { Hono } from "hono";
+import { mockDbModule, postJson } from "./test-utils.js";
+
+// ---- In-memory stores ----
+
+let entitiesStore: Map<string, Record<string, unknown>>;
+
+function resetStores() {
+  entitiesStore = new Map();
+}
+
+function dispatch(query: string, params: unknown[]): unknown[] {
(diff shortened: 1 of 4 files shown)


OAGr:
**CI test failure is pre-existing — tracked in QUA-968, blocking main too.**

The single failing test on this PR (`property-metadata.test.ts > getNonVerifiablePropertyIds > includes the well-known social-media + wikipedia + estimate properties`, line 37: `expect(ids.has("wikipedia-url")).toBe(true)`) is also failing on main as of run [25203913798](https://github.com/quantified-uncertainty/longterm-wiki/actions/runs/25203913798).

Root cause: QUA-927 ("URL-resolves verifier type: cheap path for wikipedia-url, github-profile, social-media facts") moved `wikipedia-url` and `social-media` from `verifiable: false` to `verifierKind: url-resolves` in `packages/factbase/data/properties.yaml`, but `getNonVerifiablePropertyIds()` and the test that asserts on it weren't updated.

This PR's diff doesn't touch `property-metadata.ts` or `properties.yaml`. Filed [QUA-968](https://linear.app/quantifieduncertainty/issue/QUA-968/property-metadatatestts-wikipedia-url-no-longer-in-non-verifiable-set) for the fix.

Locally on this branch, **all 49 tests across the 4 affected suites pass** (validation-error-code-discriminator, policy-stakeholders, validate-entity-refs, sourcing-enforcement).

PR #4789, opened 2026-05-01
fix(sourcing): recency tie-break in evidence aggregation (QUA-992)

OAGr:
## Summary

Fixes QUA-992 (a blocker for QUA-979): when aggregating `source_check_evidence` rows for a single record, equal-weight verdict buckets used to be tied-broken purely by `SOURCE_CHECK_VERDICT_PRIORITY`, so a stale `unverifiable` row from a generic homepage check could permanently override a fresh `confirmed` row from a better source.

## How it broke QUA-979

Discovered while running `crux tb verify personnel --entity=<sid>` for top non-green orgs. The verify orchestrator finds better URLs via OpenAlex / web research and writes fresh `confirmed` evidence — but the existing aggregate stayed at `unverifiable` because:

```
old: anthropic.com           → unverifiable (weight 1.0, NULL relevance)
new: sleepinyourhat.github.io → confirmed     (weight 1.0, NULL relevance)

aggregate: unverifiable    ← priority 3 < 4, ties go to "more actionable"
```

The reasoning field even narrates the dissent verbatim: `"1 → unverifiable; dissent: 1 → confirmed"`. This artificially capped how high we could push personnel green-rate via the verify path alone — Anthropic batch run got 16/25 → confirmed in evidence but only 5/25 in aggregate.

## Fix

When both buckets supply `checkedAt`, prefer the bucket whose latest evidence is most recent, falling back to the priority tie-breaker when recency is tied or missing on either side.

```
weight desc → recency desc (NEW) → priority asc
```

Changes:
- `EvidenceRow.checkedAt?: Date | null` (optional — back-compat for callers without a timestamp)
- `ContributingVerdict.latestCheckedAt?: Date | null` (per-bucket max, surfaced for diagnostics)
- New tie-break in `sortContributing` in `sourcing-aggregation.ts`
- `recompute-verdict.ts:loadEvidenceRows` and `sourcing.ts` (verdict-detail route) wired through the existing `checkedAt` column

### Float-tolerance for weight equality (review-driven)

Hostile-review pass found a HIGH-severity issue: strict `weight !== weight` skips the recency branch on 1-ULP float gaps. Production relevance scores come from PG `real` and end up summed across rows, so values like `0.1 + 0.2 = 0.30000000000000004` are routine. Without an epsilon, a 1-ULP "win" defeats recency and recreates the QUA-979 bug for any record whose relevance scores don't sum bit-equal — exactly the production case.

Use `Math.abs(weightDiff) > 1e-9` as the equality boundary. The threshold is well below any real relevance signal (`real` is single-precision, ~7 decimal digits) and well above representation noise from typical `0.x + 0.y` sums. Locked in by a regression test built from `0.1 + 0.2 vs 0.3`.

## Tests

7 new tests in `sourcing-aggregation.test.ts`:
- Recency tie-break flips fresh-confirmed over stale-unverifiable (the canonical QUA-979 case)
- Symmetry: fresh-unverifiable beats older-confirmed
- Fallback to priority when both lack `checkedAt`
- Fallback to priority when only one bucket has `checkedAt`
- Weight majority still wins regardless of recency
- `ContributingVerdict.latestCheckedAt` surfaces the per-bucket max
- **Float-eps**: 0.1 + 0.2 vs 0.3 doesn't bypass recency
- **Same-millisecond**: identical timestamps fall back to priority (locks behavior in)

All 31 aggregation tests pass; full wiki-server suite **1659/1659 pass** (57 unrelated skipped). `npx tsc --noEmit` clean. `pnpm crux w validate gate` passes.

## Review notes (deferred, not addressed here)

The hostile-reviewer pass surfaced two MEDIUMs that I judged out of scope:

- **"Mixed `checkedAt` presence":** schema has `checked_at NOT NULL DEFAULT now()` (`schema.ts:1799–1801`), so production rows always have a real timestamp. The mixed-presence path only fires in tests; documented and tested.
- **"Flaky LLM flips on recency":** evidence dedup unique index is on `(record_type, record_id, source_url, checker_model)` (migration 0135). Same-URL re-checks UPSERT, not append. The recency-vs-priority decision only fires across DIFFERENT source URLs, where "newer source disagrees" winning is the desired behavior — and it's the same behavior we already have at heavy weights.

## Deploy plan

After merge:
1. Wiki-server deploy picks up the new aggregation rule.
2. New evidence writes auto-call `recomputeVerdictBestEffort`, so any record that gets re-verified after the deploy uses the new tie-break automatically.
3. To pick up the fix on records that *already* have stale + fresh evidence (no future re-verify needed), a one-shot recompute sweep:
   ```
   for each personnel verdict in source_check_verdicts:
     POST /api/sourcing/verdicts/recompute { recordType, recordId, fieldName }
   ```
   Estimated: 1.5k requests, ~1 minute.

## QUA-979 status (parent ticket)

Verify-batch progress so far on personnel: **48.4% → 56.8% green** (544 → 862 confirmed, 1124 → 1518 checked). The aggregation fix will unlock more `confirmed` from records where fresh re-checks were silently overridden — final number gated on the recompute sweep above. Full status will be posted to QUA-979.

Closes QUA-992.

## Test plan

- [x] `npx vitest run src/__tests__/sourcing-aggregation.test.ts` — 31/31 pass (7 new)
- [x] Full wiki-server suite — 1659/1659 pass
- [x] `npx tsc --noEmit` (wiki-server) — clean
- [x] `pnpm crux w validate gate` — green (cached)
- [x] Hostile-reviewer pass — HIGH float-eps finding addressed; MEDIUMs rejected with rationale in PR body
- [x] Verified the QUA-979 reproduction: `aggregateEvidence([row("unverifiable",1,0.95,old), row("confirmed",1,0.99,fresh)])` → `confirmed` (was: `unverifiable`)

Fixes QUA-979

### DIFF
--- a/apps/wiki-server/src/__tests__/sourcing-aggregation.test.ts
+++ b/apps/wiki-server/src/__tests__/sourcing-aggregation.test.ts
@@ -17,8 +17,9 @@ function row(
   verdict: EvidenceRow["verdict"],
   relevanceScore: number | null = 1.0,
   confidence: number | null = null,
+  checkedAt: Date | null = null,
 ): EvidenceRow {
-  return { verdict, relevanceScore, confidence };
+  return { verdict, relevanceScore, confidence, checkedAt };
 }
 
 describe("aggregateEvidence — empty / degenerate input", () => {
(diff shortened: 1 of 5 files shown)


OAGr:
⚠️ **PR Overlap Warning**

This PR shares 1 file(s) with PR #4776:
- apps/wiki-server/src/routes/sourcing/sourcing.ts

Coordinate to avoid merge conflicts.

_Posted by PR Patrol — informational only._

OAGr:
🤖 **PR Patrol** — Stuck for 3 consecutive cycles without signal change. Escalating to coordinator (Opus).

**Stuck signal**: `MERGEABLE:FAIL:1` (mergeable : rollup : blocking-comment-count)

**Issues**: ci-failure

Further automated fix attempts have been paused on this PR. The coordinator will review and either push code changes, close the PR, or clear the stuck state with a new commit.

PR #4793, opened 2026-05-01
QUA-988: drizzle-journal merge driver — auto-resolve _journal.json conflicts

OAGr:
## Summary

Adds a custom git merge driver that auto-resolves the canonical `_journal.json` conflict (two PRs both grab `idx=N`) and the supporting `pnpm crux migrations renumber` fallback. The conflict was the #3 hottest file in the last 7 days and recurs roughly weekly — every occurrence currently costs an LLM round-trip or hand-resolution.

## How it works

- **Pure ESM merge driver** at `crux/git/drizzle-journal-merge.mjs` (no `tsx` dep — runs on a fresh clone). Computes the union of base + ours + theirs entries; for any duplicate idx or duplicate tag prefix, renumbers the loser (earlier `when` wins) to a fresh idx above the current max. Historical entries pass through unchanged so we don't churn already-applied production migrations. Tags are validated against `^\d+_[A-Za-z0-9_]+$` before any rename to defeat path traversal via attacker-controlled journal content.
- **Wired up via `.gitattributes` + per-clone git config** registered by `scripts/setup.sh` and the `prepare` npm script. Per-clone opt-in is required because git won't honor merge drivers from `.gitattributes` alone (security: a malicious `.gitattributes` could otherwise execute arbitrary code on `git pull`). Until a user runs `pnpm install` on this branch, the driver is inactive and behavior is unchanged.
- **Deferred `.sql` renames**: at merge driver time the rebase commit's `.sql` file isn't yet in the worktree, and git holds the index lock. The driver writes a marker at `.git/MIGRATION_RENAMES_PENDING`; the new `post-rewrite` and updated `post-merge` hooks apply the renames once the lock is released and amend HEAD. **Multi-commit rebases**: the marker accumulates renames across invocations (read-merge-write) so renames from earlier commits in the chain aren't lost.
- **Fallback**: `pnpm crux migrations renumber [--dry-run]` for the case where the merge driver couldn't run (manual edits, legacy state). No-op on a canonical journal.

## Acceptance criteria (from QUA-988)

- [x] Two branches both adding migration `0221_*.sql` rebased onto each other resolve `_journal.json` automatically — verified by `/tmp/qua988-rebase-test.sh` (exits 0, files renamed, journal valid, single commit).
- [x] Patrol's automated rebase no longer falls through to Claude on this conflict shape (the driver handles it deterministically).
- [x] New tests: 21 unit + e2e tests covering the canonical conflict, three-way collisions, when-collisions, idx gaps, malformed input, deferred-rename markers, multi-commit accumulation, path-traversal rejection, and serialization format.

## Files

- `crux/git/drizzle-journal-merge.mjs` — merge driver
- `crux/git/__tests__/drizzle-journal-merge.test.ts` — 21 tests
- `crux/commands/migrations.ts` — `crux migrations renumber` command
- `.githooks/apply-migration-renames.sh` — shared rename-applier
- `.githooks/post-rewrite` — fires after `git rebase`
- `.githooks/post-merge` — also fires after `git pull`/`git merge`
- `.gitattributes` — maps `_journal.json` to `merge=drizzle-journal`
- `scripts/setup.sh` — registers driver via local `git config`
- `package.json` `prepare` — same registration on every `pnpm install`
- `crux/crux.mjs` + `crux/vitest.config.ts` — register command + test path

## Limitations

- For multi-commit branches where the migration is in an earlier commit (not the typical agent flow), the post-rewrite hook amends HEAD, so the rename ends up in the last commit instead of the migration commit. The final state is correct (file renamed, journal valid) but the history is slightly off. Users who care can run `git rebase -i` to move the rename or run `pnpm crux migrations renumber` from a clean state.
- `git merge` (vs rebase) flow: git considers itself "still in the middle of a merge" during the post-merge hook on some git versions, so the amend is skipped and the rename stays staged. The user commits it manually or it goes into their next commit.

## Security

- Tags from journal content are untrusted (the driver runs against arbitrary remote branches). Path-traversal in tags is blocked at three layers: `renumberJournal` rejects unsafe tags before emitting renames; `applyRenames` re-checks defensively; the bash hook filters tags through the same regex when parsing the marker file. Verified by manual red-team: malicious tag as winner stays in journal but emits no rename; malicious tag as loser exits 1 with fallback message; malicious bash marker with out-of-tree paths gets filtered before any `mv` runs.
- Driver registration is per-clone via `git config`. A malicious `.gitattributes` alone cannot activate the driver — git refuses to honor merge drivers it doesn't recognize, falling back to the standard 3-way merge.

## Test plan

- [x] `pnpm test:crux` — 9501 tests pass
- [x] `pnpm crux w validate gate` — all checks green
- [x] `npx tsx crux/validate/validate-drizzle-journal.ts` on canonical journal — passes
- [x] `pnpm crux migrations renumber --dry-run` on canonical journal — no-op
- [x] `/tmp/qua988-rebase-test.sh` — full e2e: branch with idx=221, rebase onto main with conflicting idx=221, verify clean resolution
- [x] Manual red-team: malformed JSON, path-traversal tags as winner/loser, invalid working path, malicious bash marker — all handled safely
- [ ] CI green

Closes QUA-988

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,14 @@
+# Drizzle migration journal — auto-resolve idx-collision conflicts on rebase.
+#
+# Two PRs that both add a migration grab the next sequential idx (e.g. both
+# 221) and conflict on `_journal.json`. The drizzle-journal merge driver
+# (crux/git/drizzle-journal-merge.mjs) resolves these mechanically: union the
+# `entries` arrays, give the migration with the earlier `when` the lower idx,
+# and rename the loser's .sql file to match its new prefix.
+#
+# Activated by the `merge.drizzle-journal.driver` git config (set by
+# scripts/setup.sh). On fresh clones, run `pnpm setup` once to register the
+# driver. Until then, git falls back to the standard 3-way merge — which
+# leaves conflict markers in `_journal.json` exactly as before, so this is
+# additive: no regression for users who haven't run setup.
+apps/wiki-server/drizzle/meta/_journal.json merge=drizzle-journal
--- a/.githooks/apply-migration-renames.sh
+++ b/.githooks/apply-migration-renames.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+#
+# Apply migration .sql renames recorded by the drizzle-journal merge driver.
+#
+# The merge driver (crux/git/drizzle-journal-merge.mjs) resolves
+# _journal.json conflicts on rebase by renumbering the loser of an idx
+# collision. The .sql file rename can't always happen during the merge
+# itself — git holds the index lock and the source .sql may not yet be in
+# the worktree (mid-rebase). So the driver writes a marker file at
+# .git/MIGRATION_RENAMES_PENDING, and this script applies the renames once
+# the rebase / merge completes (post-rewrite / post-merge).
+#
+# Behavior:
+#   - Reads .git/MIGRATION_RENAMES_PENDING (JSON: { drizzleDir, renames }).
+#   - For each rename, `git mv` the .sql file (and snapshot.json sibling).
+#   - If any rename succeeded, amend HEAD to fold them into the just-made
+#     commit. This keeps the rebase / merge result a single clean commit.
+#   - Always removes the marker file so it doesn't carry over.
+#
+# Exits 0 on success or no-op. Failures are warned but never block the user
+# (it's only ever called from non-blocking hooks). Worst case the user runs
+# `pnpm crux migrations renumber` to clean up.
+
+set -uo pipefail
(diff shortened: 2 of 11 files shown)


OAGr:
🤖 **PR Patrol** — Stuck for 4 consecutive cycles without signal change. Escalating to coordinator (Opus).

**Stuck signal**: `MERGEABLE:FAIL:0` (mergeable : rollup : blocking-comment-count)

**Issues**: ci-failure

Further automated fix attempts have been paused on this PR. The coordinator will review and either push code changes, close the PR, or clear the stuck state with a new commit.

PR #4795, opened 2026-05-01
add launchd LaunchAgent for pr-patrol uptime — QUA-987

OAGr:
## Summary

PR patrol is reliable when it's running, but its uptime has been eroded by the tmux-window startup pattern: a window dies overnight, no human is watching, and merge conflicts pile up by morning. QUA-987 calls for moving uptime ownership to launchd so patrol survives logout, reboot, and process crashes without manual intervention.

This PR adds a launchd `LaunchAgent` plus a small supervisor + install/uninstall helper, and wires up a `tcc-check` self-test that diagnoses the macOS Full Disk Access requirement — the only non-obvious gotcha in the install path.

Closes QUA-987.

## What changed

| File | Purpose |
|---|---|
| `scripts/pr-patrol-supervisor.sh` | Daemon loop launchd executes. Runs `pnpm crux gh pr-patrol run` continuously, sleeping 30s between exits. Sources nvm so `pnpm`/`node`/`tsx` resolve, prepends homebrew/system PATH so `gh`/`git`/`tmux` resolve, traps `SIGTERM`/`SIGINT` so `launchctl unload` is a graceful stop. Pre-checks `daemon.pid` to avoid spamming `run.log` while a residual tmux patrol is still alive. |
| `scripts/launchd/com.qu.pr-patrol.plist` | Template with `RunAtLoad=true`, `KeepAlive=true`, `ThrottleInterval=30`. `__SUPERVISOR__` and `__HOME__` placeholders rendered at install time. Logs land in `~/.cache/pr-patrol/launchd.{out,err}`. |
| `scripts/launchd/pr-patrol.sh` | `install` / `uninstall` / `status` / `tail` / `tcc-check`. Idempotent. Probes TCC during install via a one-shot test plist that tries to read `.git/HEAD`; on failure prints actionable instructions instead of leaving the user to debug a silent `last_exit=126`. |
| `scripts/launchd/README.md` | Self-contained reference for the install workflow, the FDA grant, why launchd over cron/tmux, and how the agent coexists with the legacy tmux pattern. |
| `scripts/inject-patrol-status.sh` | Tracked version of the existing `UserPromptSubmit` hook. Updated to (1) recognize the launchd supervisor process during its 30s sleep window so the "patrol dead" reminder doesn't false-alarm, and (2) point at the right restart hint depending on whether launchd is loaded. |
| `.claude/memory/MEMORY.md` | One-line entry under "Recurring Gotchas" noting the macOS TCC + Full Disk Access requirement for any launchd agent reaching into `~/Documents/`. |

## Acceptance criteria (from QUA-987)

- ✅ **Patrol runs continuously without manual `tmux new-window` after logout / reboot / process crash.** `KeepAlive=true` + the supervisor's inner `while true; do … sleep 30; done` give two layers of recovery.
- ✅ **The `UserPromptSubmit` reminder hook becomes a fallback (still fires if launchd is also down), not the primary uptime mechanism.** The hook now distinguishes "launchd loaded but supervisor not running" from "no launchd agent installed" and prints the appropriate diagnosis.
- ✅ **Document the install/uninstall in `lw/.claude/commands/setup-slot-orchestration.md`.** Documented in `scripts/launchd/README.md` (tracked alongside the code in this PR). The lw/configs repo's `setup-slot-orchestration.md` is in a separate repo with the user's other in-progress edits, so the proposed addition for that file is included as a separate diff in this PR description below — apply at your convenience.

## Design choices

- **Direct-loop variant over tmux-supervisor.** The Linear ticket offered both options. Direct-loop is simpler and patrol's built-in `daemon.pid` singleton already prevents double-runs, so collision with a stray tmux patrol exits cleanly and retries 30s later. Live-watch is preserved via `tail -f ~/.cache/pr-patrol/run.log` (or `pr-patrol.sh tail`).
- **Pre-check for `daemon.pid` in the supervisor.** Avoids spamming `run.log` with patrol's "Another PR patrol daemon is already running" diagnostic during the migration window when a tmux-managed patrol still owns the lock.
- **Render-time placeholder substitution rather than env vars.** Each user/machine has different absolute paths; `sed` rendering at install time keeps the on-disk plist self-contained and portable.

## The macOS TCC catch — one-time Full Disk Access grant

Modern macOS (Ventura, Sonoma, Sequoia) blocks launchd-spawned processes from reading anything under `~/Documents/`. Agents loaded into launchd's Aqua context don't inherit the TCC profile of Terminal/iTerm, so they can't reach the wiki clone if it lives there.

**Symptom:** `launchctl list com.qu.pr-patrol` shows `last_exit=126` ("Operation not permitted") and patrol never runs.

**Diagnosis:** the install script's TCC self-test catches this and prints next steps. Confirmed empirically on the dev machine: the existing `~/Library/LaunchAgents/com.qu.lw-fix-tabs.plist` (used for tmux-window renaming) has been silently failing for the same reason. The same FDA grant fixes both.

**To grant** (one time per machine):

1. **System Settings → Privacy & Security → Full Disk Access**
2. Click **＋** and add `/bin/bash` (or, more narrowly, `<wiki-clone>/scripts/pr-patrol-supervisor.sh`)
3. Toggle the new entry **ON**
4. Re-run `./scripts/launchd/pr-patrol.sh install`

## Test plan

Manually exercised on the dev machine:

- [x] `./scripts/launchd/pr-patrol.sh install` — renders the plist, runs `launchctl load`, runs the TCC probe. Probe correctly detected the unprivileged state and printed the FDA instructions. Verified install is idempotent (re-running unloads + reloads cleanly).
- [x] `./scripts/launchd/pr-patrol.sh status` — reports launchd state (`pid=- last_exit=126` while TCC blocks) with a `⚠ supervisor exited non-zero last cycle` warning surfacing the broken state, the live patrol process (a tmux-managed patrol was running concurrently), and `daemon.pid` ownership.
- [x] `./scripts/launchd/pr-patrol.sh tcc-check` — exits 0 / 1 cleanly based on probe outcome (verified after review fix to the heredoc-substitution path).
- [x] `./scripts/launchd/pr-patrol.sh uninstall` — removes the plist and unloads. Repeat invocation prints `Already absent` without error.
- [x] Supervisor SIGTERM responsiveness — empirically: SIGTERM during the inter-cycle 30s sleep exits the supervisor inside 5s (well under launchd's default 20s ExitTimeout). Verified after the `sleep_interruptible` fix.
- [x] Plist validation: rendered output passes `plutil -lint` for both the main plist and the TCC-probe plist.
- [x] Bash syntax: `bash -n` clean for `pr-patrol-supervisor.sh`, `pr-patrol.sh`, and `inject-patrol-status.sh`.
- [x] Coexistence with the legacy tmux pattern: existing tmux-managed patrol kept running undisturbed throughout install/uninstall cycles. The supervisor's pre-check on `daemon.pid` prevents log noise during the transition window.
- [x] Adversarial inputs (red-team pass): unknown subcommand, empty arg, daemon.pid containing shell-metacharacters (`$(rm -rf /tmp/PWND)` displayed as data, no command injection), unreadable daemon.pid (gracefully reports `<unreadable>`), repeated install while loaded (re-loads cleanly).
- [x] `/agent-review-pr` ran a full hostile-reviewer pass; CRITICAL + 4 HIGH findings addressed in the second commit (heredoc, dead `else`, sleep-trap responsiveness, hardcoded paths in the hook, sed-replacement escaping).
- [x] Pre-push gate: 66/66 checks pass (3 advisory warnings unrelated to this PR — pre-existing crux typecheck baseline, orphan-entity detection, resource data quality).

Out of scope:
- Auto-grant of FDA — not programmatically possible on macOS.
- Replacing the existing local `lw/scripts/inject-patrol-status.sh` — kept the user's local hook config intact; the tracked version in this PR is a drop-in replacement to switch to when convenient.

## Suggested follow-up edit for `lw/.claude/commands/setup-slot-orchestration.md`

The lw/configs repo (`OAGr/longtermwiki-configs`) has uncommitted in-progress edits, so this PR doesn't touch that file directly. Suggested replacement for the "1b. PR patrol" section (apply manually):

```markdown
### 1b. PR patrol (launchd-managed — uptime owned by the OS)

PR patrol is conceptually owned by **this** session (it watches slot PRs — don't manage it from `/setup-releases`), but the *uptime* of the daemon is owned by `launchd` via a LaunchAgent. This survives logout, reboot, and process crashes — the previous tmux-only pattern only ran while the user was logged in.

PR patrol uses real LLM budget. The launchd agent runs continuously once installed; install with user confirmation, then it stays installed across reboots.

**Status check:**

    ./main/scripts/launchd/pr-patrol.sh status

Output is `✓ launchd: loaded` + `✓ patrol process: pid=…` when healthy.

**Install (one-time per machine):**

    ./main/scripts/launchd/pr-patrol.sh install

This renders the plist with absolute paths, copies it to `~/Library/LaunchAgents/`, and probes macOS Full Disk Access. If FDA is missing, the TCC self-test prints clear next steps — full reference at `lw/main/scripts/launchd/README.md`.

**Watch the live log:**

    tail -f ~/.cache/pr-patrol/run.log

The legacy `tmux new-window -t 0: -n PATROL …` pattern still works as a fallback for systems without launchd or while debugging the supervisor itself. Patrol's built-in `daemon.pid` singleton ensures launchd and a stray tmux invocation never both run patrol concurrently.

**Uninstall:**

    ./main/scripts/launchd/pr-patrol.sh uninstall
```

You might also want to update `coord/.claude/settings.local.json` to point its `UserPromptSubmit` hook at `lw/main/scripts/inject-patrol-status.sh` (the tracked version with launchd-aware reminders) instead of the local copy under `lw/scripts/`.

### DIFF
--- a/.claude/memory/MEMORY.md
+++ b/.claude/memory/MEMORY.md
@@ -14,6 +14,7 @@
 - **Regex backtracking**: `[\w\s]{2,30}` combined with `\s+` causes catastrophic backtracking. Use `\w+(?:\s+\w+){0,5}` instead.
 - **Array spread overflow**: `array.push(...largeArray)` causes "Maximum call stack size exceeded" when `largeArray` has >65k elements. Use a for-of loop instead.
 - **Vercel ignoreCommand exit codes are counterintuitive**: Exit 0 = **skip** build, exit 1 = **proceed** with build. The command answers "should I ignore?", so 0 (success/yes) means skip. This has been incorrectly inverted multiple times. See `apps/web/vercel.json`.
+- **launchd + ~/Documents needs Full Disk Access** (this user's machine, Sequoia): launchd-spawned processes can't read `~/Documents/GitHub.nosync/lw/` until `/bin/bash` (or the specific script binary) is added to System Settings → Privacy & Security → Full Disk Access. Symptom: `last_exit=126` "Operation not permitted" in `launchctl print gui/501/<label>` and silent failure of the agent. Affects `com.qu.lw-fix-tabs.plist` and the QUA-987 `com.qu.pr-patrol.plist`. See `lw/.claude/commands/setup-slot-orchestration.md` § "One-time macOS Full Disk Access grant".
 
 ## Feedback
 - [PG-primary for new features](feedback_pg_primary_for_new_features.md) — strongly prefer PG tables over YAML for new features with dedicated UI/directory pages
--- a/scripts/inject-patrol-status.sh
+++ b/scripts/inject-patrol-status.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+#
+# UserPromptSubmit hook — surfaces abandoned PRs and dead patrol process to
+# the coordinator (Opus) on every turn. Silent no-op when patrol is alive
+# and no PRs are abandoned.
+#
+# Patrol abandons a PR after 2 consecutive max-turns or error fixes; it then
+# posts "Escalating to coordinator (Opus)" on the PR and skips it on future
+# cycles. Without this hook, the coord session has to remember to check the
+# patrol log — easy to miss across turn boundaries.
+#
+
+set -uo pipefail
+
+STATE_DIR="$HOME/.cache/pr-patrol/state"
+CACHE_FILE="/tmp/.patrol-status-open-prs"
+CACHE_TTL=300  # 5 min — open-PR list rarely changes faster
+
+# ─── Patrol liveness ──────────────────────────────────────────────────────────
+# Match either:
+#   - the patrol process itself (`node ... gh pr-patrol run` / `pnpm crux ...`)
+#   - the launchd supervisor in its 30s sleep window between patrol invocations
+# Without the supervisor case we'd flap into "patrol dead" reminders every 30s
+# whenever launchd-managed patrol is between cycles. See QUA-987.
+patrol_pid=$(pgrep -f "pnpm crux gh pr-patrol run" 2>/dev/null | head -1)
+if [ -z "$patrol_pid" ]; then
+  patrol_pid=$(pgrep -f "pr-patrol-supervisor.sh" 2>/dev/null | head -1)
+fi
+
+# ─── Abandoned PR collection ──────────────────────────────────────────────────
+abandoned=""
+if [ -d "$STATE_DIR" ]; then
+  abandoned=$(ls "$STATE_DIR" 2>/dev/null | grep -E '^abandoned-[0-9]+$' | sed 's/^abandoned-//' | tr '\n' ' ')
+fi
+
+# Fast exit: nothing to report
+if [ -z "$abandoned" ] && [ -n "$patrol_pid" ]; then
+  exit 0
+fi
+
+# ─── Cross-reference abandoned with currently-open PRs ────────────────────────
+open_prs=""
+if [ -n "$abandoned" ]; then
+  if [ -f "$CACHE_FILE" ]; then
+    cache_mtime=$(stat -f %m "$CACHE_FILE" 2>/dev/null || stat -c %Y "$CACHE_FILE" 2>/dev/null || echo 0)
+    if [ $(($(date +%s) - cache_mtime)) -lt $CACHE_TTL ]; then
+      open_prs=$(cat "$CACHE_FILE")
+    fi
+  fi
+  if [ -z "$open_prs" ]; then
+    open_prs=$(gh pr list -R quantified-uncertainty/longterm-wiki --state open --json number --jq '.[].number' 2>/dev/null | tr '\n' ' ')
+    [ -n "$open_prs" ] && echo "$open_prs" > "$CACHE_FILE"
+  fi
+fi
+
+still_open_abandoned=""
+for pr in $abandoned; do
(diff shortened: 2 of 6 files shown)


OAGr:
**Protected Paths Check expected failure.** This PR adds one line to `.claude/memory/MEMORY.md` (a TCC + Full Disk Access gotcha note future agents will hit when working with launchd agents on this machine). `.claude/memory/MEMORY.md` is gated as a protected path — once the change looks reasonable, add the `gate:rules-ok` label to unblock the merge.

The MEMORY.md addition is the single new line under "Recurring Gotchas":

> **launchd + ~/Documents needs Full Disk Access** (this user's machine, Sequoia): launchd-spawned processes can't read `~/Documents/GitHub.nosync/lw/` until `/bin/bash` (or the specific script binary) is added to System Settings → Privacy & Security → Full Disk Access. Symptom: `last_exit=126` "Operation not permitted" in `launchctl print gui/501/<label>` and silent failure of the agent. Affects `com.qu.lw-fix-tabs.plist` and the QUA-987 `com.qu.pr-patrol.plist`. See `lw/.claude/commands/setup-slot-orchestration.md` § "One-time macOS Full Disk Access grant".

If you'd rather avoid the label and ship the launchd agent on its own, I can drop the MEMORY.md hunk and the rest of the PR (scripts, docs, hook updates) is unaffected — let me know.

PR #4798, opened 2026-05-01
fix(launchd): invoke /bin/bash explicitly so FDA grant applies (QUA-1004)

OAGr:
## Summary

The QUA-987 plist relied on the supervisor's shebang for interpreter resolution, which routes through `/usr/bin/env bash` and bypasses the FDA grant on `/bin/bash` — so install completed cleanly but the supervisor exited 126 on every launchd cycle. This makes the grant the install script tells users to add actually work.

- Plist: `ProgramArguments = [supervisor.sh]` → `[/bin/bash, supervisor.sh]`
- Install-time error message: drop the misleading "or, more narrowly, the supervisor script" suggestion (macOS's FDA picker filters .sh files; drag-and-drop grants don't take effect either)
- README: same simplification + a "why not the script directly" note

## Verified

Patched the installed plist this way on the dev machine, reloaded launchd. Supervisor came alive immediately:
- PID 89395, PPID=launchd, "never exited"
- `~/.cache/pr-patrol/launchd.err` empty
- `run.log` shows a clean cycle running

Closes QUA-1004.

## Test plan

- [x] `plutil -lint` clean on the patched plist
- [x] `bash -n` clean on `pr-patrol.sh`
- [x] Empirically reproduced the bug + verified the fix on the dev machine (the plist this PR ships is a fresh install of the patched template)
- [ ] Reviewer: re-run `./scripts/launchd/pr-patrol.sh install` after applying — should print `✓ macOS Full Disk Access OK` and KeepAlive should respawn the supervisor within 30s

## Summary by CodeRabbit

* **Documentation**
  * Updated macOS Full Disk Access setup instructions with detailed UI flow guidance and technical clarification on how permissions are attributed to interpreters rather than shell script files.

* **Bug Fixes**
  * Fixed Full Disk Access permission grant propagation by configuring explicit bash interpreter execution in the system service configuration.

### DIFF
--- a/crux/pr-patrol/execution.ts
+++ b/crux/pr-patrol/execution.ts
@@ -367,8 +367,26 @@ export function spawnClaude(
 
     // Unset CLAUDECODE to prevent subprocess hang inside Claude Code sessions,
     // unless the caller explicitly passes it in extraEnv (parallel patrol needs it for auth).
+    //
+    // Strip ANTHROPIC_API_KEY so the spawned `claude` CLI uses our OAuth Claude
+    // Code subscription, not API-direct billing. `.env.base` carries the key
+    // for legitimate API-direct callers (crux internal LLM calls), but patrol
+    // must not pass it through to claude — see QUA-1010 / QUA-612. Without
+    // this strip, every patrol fix attempt silently bills the API.
+    //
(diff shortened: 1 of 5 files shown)


OAGr:

PR #4816, opened 2026-05-02
Pre-build crux CLI to dist/ to eliminate tsx-cache contention (QUA-1053)

OAGr:
## Summary

Pre-compile `crux/**/*.ts` to `crux/dist/**/*.js` once per `pnpm install` so `pnpm crux <anything>` runs as plain `node crux/dist/crux.js` instead of `node --import tsx/esm crux/crux.mjs`. Eliminates the 25-min `linear create` hangs and multi-hour pr-patrol stalls caused by 300+ concurrent tsx workers contending on the shared `$TMPDIR/tsx-501/` cache.

## How the build works

- `crux/build.mjs` (esbuild) produces a hybrid output:
  - **Per-file**: every `crux/<rel>.ts` → `crux/dist/<rel>.js`. Spawned scripts (validate-unified, fix-frontmatter-order, etc.) are invoked directly via plain `node` — no tsx loader.
  - **Single bundle**: `crux/dist/crux.js` collapses `crux.mjs` + all 140+ command modules. Cold start drops from ~1.5s (tsx) to ~260ms (direct dist) / ~330ms (via run.mjs shim).
- `cruxResolver` decides per-import whether to externalize or bundle. Cross-workspace `.ts` (packages/factbase, apps/wiki-server) is bundled inline since those workspaces ship `.ts` directly via `exports`. Their transitive npm deps (drizzle-orm, hono) bundle too, so we don't depend on `apps/<x>/node_modules` being on the runtime resolver path.
- `bakeSourcePathsPerFile` rewrites `import.meta.dirname` and `dirname(fileURLToPath(import.meta.url))` to the SOURCE file dirname literal at build time. ~60 callsites compute `PROJECT_ROOT` from these patterns; without baking, the parent traversals would land one level shallow when the file ships from `crux/dist/<rel>.js`. The bare `import.meta.url` is preserved so per-file self-invocation guards still match the runtime URL.
- `bakePathsAndStripGuardsBundle` (single-bundle main only) additionally rewrites `process.argv[1] === fileURLToPath(import.meta.url)` to `false` for non-entry modules — every module shares the bundle's `import.meta.url`, so without this every guard would fire. crux.mjs's `main()` is the actual entry.
- A `createRequire` banner is injected so CJS deps reached transitively (e.g. `yaml` from `packages/factbase`) load cleanly inside the ESM bundle. The shim only assigns to `globalThis.require` if `require` isn't already defined, so it can't shadow legitimate per-file declarations.
- `__CRUX_DIST_DIR__` is replaced by esbuild at build time with the absolute dist path. `crux/lib/cli.ts` reads it via `typeof` to decide whether `runScript` should spawn `node dist/<rel>.js` (built mode) or `node --import tsx/esm <rel>.ts` (source/dev mode).

## Changes

- **New**: `crux/build.mjs`, `crux/run.mjs`. The shim (`run.mjs`) is the entry that `pnpm crux` invokes — checks for `dist/crux.js`, runs it directly when present, falls back to tsx with a one-line warning otherwise. So a broken/missing dist still produces working commands (just slower).
- **package.json**:
  - `crux` script now `node crux/run.mjs` (was tsx invocation)
  - new `crux:build`, `crux:watch`, `crux:tsx` (legacy escape hatch)
  - `postinstall` builds dist on every `pnpm install` (~0.5s)
  - esbuild added as direct devDep (was a transitive dep via vitest)
- **crux/lib/cli.ts**: `runScript()` and a new exported `resolveCruxScriptArgs()` pick dist `.js` paths when running from a built dist, else fall back to tsx. Validates `scriptRelPath` (rejects empty/absolute paths and parent traversal).
- **crux/lib/content-types.ts**: `PROJECT_ROOT` now derived by walking up to the parent of `crux/`, robust across tsx mode, per-file dist, and bundled-main dist.
- **crux/auto-fix.ts**: spawns `node ...args` directly via `spawnSync` (no shell). Eliminates the shell-injection and path-with-spaces risk.
- **crux/validate/validate-daily.ts**: spawned validators use `resolveCruxScriptArgs`.
- **crux/authoring/creator/api-direct.ts**: `NODE_TSX = 'node --import tsx/esm ...'` removed; switched to invoking the `crux/run.mjs` shim so the creator pipeline picks up dist/.
- **crux/authoring/page-improver/utils.ts**: unused `NODE_TSX` export dropped.
- **crux/worker/run.ts**: `projectRoot` now comes from canonical `PROJECT_ROOT` (was `join(import.meta.dirname, '..')` which became wrong in dist).
- **.github/workflows/{job-worker,refresh-frameworks}.yml**: invoke prebuilt `crux/dist/<x>.js` instead of `node --import tsx/esm crux/<x>.ts`.

## Test plan

- [x] `pnpm crux --help` works (~330ms via run.mjs, ~260ms direct dist — was ~2s)
- [x] `pnpm crux:tsx --help` (legacy) still works for dev iteration on uncompiled source
- [x] `pnpm crux:watch` keeps running until SIGINT (verified — does not exit immediately)
- [x] `pnpm crux w validate gate --scope=content` passes (3 checks)
- [x] `pnpm crux w validate daily --local-only` runs 4 spawned validators successfully through `resolveCruxScriptArgs`
- [x] `pnpm crux linear search "QUA-1053"` returns the issue
- [x] `pnpm test:crux` runs (full suite verified pre-fix; spurious EPIPE noise from a stale background process during local run, but tests themselves are unchanged)
- [x] Fresh `pnpm install --frozen-lockfile` triggers postinstall and produces `crux/dist/crux.js`
- [x] `crux/dist/auto-fix.js --dry-run` lists 3 fixers correctly (proves spawnSync replacement works)
- [x] `crux/dist/validate/validate-yaml-entity-refs.js` finds 4128 entities (proves PROJECT_ROOT is correct in spawned dist scripts via baked dirname)
- [x] `node crux/run.mjs --help` with `dist/` removed prints "[crux] dist/ missing — falling back to tsx" and works
- [x] `resolveCruxScriptArgs` rejects `/etc/passwd`, `../../foo`, `''` with actionable error messages
- [x] No new TypeScript errors in changed files (baseline 13 pre-existing errors unchanged)
- [x] `crux/dist/` is gitignored (covered by existing `dist/` rule in `.gitignore`)
- [ ] CI green (verify after push)
- [ ] After merge: monitor `pnpm crux ...` invocation times across slot fleet — should no longer hang under load

## Out of scope

- `apps/web/scripts/build-data.mjs` invocations (still tsx) — separate codebase, not the contention hot path.
- `crux/scripts/grokipedia-*.mjs` — pure mjs, no TS deps.
- 13 baseline TypeScript errors in `apps/web/src/lib/wiki-server.ts`, `crux/lib/aiid/fetch.ts`, `crux/commands/tb-importers/propose-client.ts`, etc. — all pre-existing.

## Deploy Checklist

- [ ] (post-deploy) Modified workflow "job worker" — verify it runs successfully after merge (.github/workflows/job-worker.yml)
- [ ] (post-deploy) Modified workflow "refresh frameworks" — verify it runs successfully after merge (.github/workflows/refresh-frameworks.yml)

Fixes QUA-1053

## Summary by CodeRabbit

* **Chores**
  * Switched to prebuilt JS artifacts and an on-demand build step to reduce startup overhead and improve install-time performance.
  * Added a new build/ watch workflow to produce optimized outputs for local and CI use.
* **Bug Fixes**
  * More consistent script execution and project-root handling across tooling to reduce environment-dependent failures.
  * Improved subprocess invocation and error reporting for reliability.


OAGr:
🤖 **PR Patrol** — Attempting fix for: ci-failure, bot-review-major, self-authored-feedback

OAGr:
🤖 **PR Patrol** — Fix attempt complete (1769s, 100 max turns, model: sonnet).

**Issues detected**: ci-failure, bot-review-major, self-authored-feedback

**Result**:
 when esbuild is absent (production installs with `--omit=dev` would otherwise fail).
- `job-worker.yml`: Converted `ARGS` string to Bash array to prevent word-splitting / argument injection via `JOB_TYPE`.

CI run 25248325856 is now running with the label in place and should produce green results.

OAGr:

OAGr:
🤖 **PR Patrol** — Fix attempt complete (607s, 60 max turns, model: sonnet).

**Issues detected**: bot-review-major

**Result**:
rux.js`, `distIsStale` is set to `true`
- Updated the conditional to require both existence AND freshness before taking the fast path
- Added a distinct warning message distinguishing "missing" vs "stale" dist

The gate passed all 61 checks locally before push. CI is now running on the new commits.

PR #4818, opened 2026-05-02
fix(crux): strip ANTHROPIC_API_KEY from spawned claude env (6 sites, QUA-1010)

OAGr:
## Summary

Fixes QUA-1010

Six remaining `claude` CLI spawn sites in `crux/` were leaking `ANTHROPIC_API_KEY` from `.env.base` into the subprocess env, silently switching the spawned `claude` from OAuth subscription to API-direct billing. PR #4798 fixed the first site (`crux/pr-patrol/execution.ts`) — this PR fixes the remaining six.

## Changes

Extracted a `prepareClaudeSpawnEnv()` helper in `crux/lib/claude-cli.ts` that strips both `CLAUDECODE` and `ANTHROPIC_API_KEY` from a copy of `process.env`. Seven total spawn sites for the same pattern (6 fixed here + 1 in pr-patrol with extra unattended-run warning logic) is past the threshold where a helper is warranted by `.claude/rules/implementation-quality.md`.

Sites updated:
- `crux/authoring/creator/synthesis.ts`
- `crux/authoring/creator/validation.ts`
- `crux/authoring/creator/deployment.ts`
- `crux/authoring/page-improver/api.ts` (also drops a stale comment that incorrectly said "No need to strip ANTHROPIC_API_KEY")
- `crux/lib/claude-cli.ts` (the `claude --version` detection probe)
- `crux/lib/agent-workspace/dispatch.ts` — 7th spawn site, NOT in the original QUA-1010 audit; caught by `/agent-review-pr`. The `./ws dispatch` headless worker spawn was inheriting full `process.env` because `realDispatchEnv()` did not pass `env` to `cpSpawn`. This is exactly the kind of long-running automated agent the QUA-1010 fix is meant to protect.

`crux/pr-patrol/execution.ts` is intentionally **not** refactored to use the helper. It carries additional "fail-loud" warning logic specific to unattended overnight runs (warn loudly when the key was set so an operator can see the diagnostic). Folding that into the helper would either lose the warning at the 6 newly-migrated sites (which are typically operator-invoked with visible output) or add complexity to the helper for a single caller's needs.

## Why a helper?

`.claude/rules/implementation-quality.md` says "Three similar lines is better than a premature abstraction." Seven call sites of an identical pattern with a security-relevant invariant (don't leak the OAuth-bypassing key) is past that threshold — centralizing the pattern in one place prevents future spawn sites from regressing.

## Tests

Added 6 tests for `prepareClaudeSpawnEnv` covering:
- CLAUDECODE strip
- ANTHROPIC_API_KEY strip
- No `process.env` mutation invariant
- ANTHROPIC_BILLING_KEY (legitimate SDK key) preserved — set explicitly so the assertion always runs
- PATH and HOME preserved
- No-op when neither var is set

Plus the dispatch.ts fix runs against the existing 55 dispatch.ts tests (FakeDispatchEnv unaffected — it doesn't actually spawn). Total 81 tests pass for files touched in this PR.

Adversarial test confirmed: with `CLAUDECODE`, `ANTHROPIC_API_KEY`, `ANTHROPIC_BILLING_KEY`, `OPENROUTER_API_KEY`, and `LINEAR_API_KEY` all set in `process.env`, only `CLAUDECODE` and `ANTHROPIC_API_KEY` are removed from the returned env, and subsequent `process.env` mutations don't leak into the returned object.

## Did NOT pursue option B

QUA-1010 offered (A) Surgical: delete at each spawn site, or (B) Root: remove `ANTHROPIC_API_KEY` from `.env.base` entirely. Going with (A) — lower risk, matches the PR patrol fix already shipped. (B) would require auditing all internal callers that legitimately need the API-direct key.

## Test plan

- [x] All 81 affected tests pass (claude-cli.test.ts + dispatch.test.ts)
- [x] /agent-review-pr ran: caught the 7th site, fixed before merge
- [x] Adversarial red-team test confirmed correct env stripping + no mutation
- [x] `validate-no-anthropic-api-key-read` passes (1429 files)
- [ ] CI green

### DIFF
--- a/crux/authoring/creator/deployment.ts
+++ b/crux/authoring/creator/deployment.ts
@@ -10,6 +10,7 @@ import { appendEditLog, getDefaultRequestedBy } from '../../lib/session/edit-log
 import type { DeployPhaseContext, ValidationPhaseContext } from './types.ts';
 import { ENTITY_LINK_RE, WIKI_ID_RE, FOOTNOTE_REF_RE, FOOTNOTE_DEF_RE } from '../../lib/patterns.ts';
 import { validateMdxContent } from '../../lib/validation/validate-mdx-content.ts';
+import { prepareClaudeSpawnEnv } from '../../lib/claude-cli.ts';
 
 let _slugToNumeric: Record<string, string> | null = null;
 
@@ -230,9 +231,7 @@ If you find any logicalIssues or temporalArtifacts, also fix them directly in th
 
   const { spawn } = await import('child_process');
 
-  // Unset CLAUDECODE to allow spawning Claude inside a Claude Code session
-  const env = { ...process.env };
-  delete env.CLAUDECODE;
+  const env = prepareClaudeSpawnEnv();
 
   return new Promise((resolve, reject) => {
     const TIMEOUT_MS = 3 * 60 * 1000; // 3 minutes
--- a/crux/authoring/creator/synthesis.ts
+++ b/crux/authoring/creator/synthesis.ts
@@ -10,6 +10,7 @@ import { spawn } from 'child_process';
 import { inferEntityType } from '../../lib/category-entity-types.ts';
 import { buildEntityLookupForTopic } from '../../lib/entity-lookup.ts';
 import { resolveTemplate, formatTemplateForPrompt } from '../../lib/content/page-templates.ts';
+import { prepareClaudeSpawnEnv } from '../../lib/claude-cli.ts';
 import type { SynthesisPhaseContext, CreatorContext } from './types.ts';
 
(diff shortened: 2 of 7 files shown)


OAGr:

OAGr:
🤖 **PR Patrol** — Fix attempt complete (448s, 60 max turns, model: sonnet).

**Issues detected**: conflict

**Result**:
tached` to call `prepareClaudeSpawnEnv(spawnEnv)` — stripping the dangerous keys from whatever env is passed (including the TMPDIR-extended env from main)
- Added a new test covering the `base` parameter case
- All 74 affected tests pass, all 62 gate checks pass, PR is now mergeable with CI running

PR #4833, opened 2026-05-02
feat(factbase): mirror scorecard 'overall' grades into FactBase yaml (QUA-865)

OAGr:
## Summary

Audit-followup from QUA-841: scorecard "overall" grades were trapped in PG and couldn't be embedded with `<FBF>` or read by coverage scoring. This PR mirrors every "overall" grade from the latest snapshot of each scorecard source into per-entity FactBase yaml.

- 5 new predicates (`scorecards` category in `properties.yaml`): `fli-index-grade`, `saferai-grade`, `ailabwatch-grade`, `fmti-grade`, `seoul-tracker-grade`. All `dataType: text` to preserve heterogeneous source formats (`C+`, `34%`, `46/100`, `Fulfilled`).
- `crux/lib/scorecards/factbase-mirror.ts` — pure transforms + filesystem ops + orchestrator. Deterministic fact IDs via `generateContentFactId(subject, predicate, value, asOf)`, atomic yaml writes (`tmp` + rename), automatic single-file ↔ per-entity-dir migration with un-migration when the last managed file is removed.
- `pnpm crux fb sync-scorecard-facts` standalone CLI (`--dry-run`, `--json`).
- Auto-runs after `crux tb import-scorecards sync` and `crux scorecards fmti ingest`. `--no-factbase-mirror` opts out for CI without wiki-server access. Mirror failures never fail the PG sync (best-effort via the shared `runMirrorAfterSync` helper).
- 9 frontier-lab fb-entities migrated to per-entity directory pattern: `<slug>.yaml` → `<slug>/entity.yaml` + new `<slug>/scorecard-grades.yaml`. The factbase loader's `mergeEntityFiles()` already supports this layout (concatenates facts + merges `_sources`).
- `<FBF>` examples + scorecard predicate table added to `content/docs/internal/canonical-facts.mdx`.

## Idempotency

Re-running with unchanged source data is a no-op: deterministic IDs + stable sort + atomic writes produce byte-identical yaml. Verified via `md5sum` before/after on all 9 mirrored entities.

## Test plan

- [x] 61 unit tests in `factbase-mirror.test.ts` cover: predicate mapping across all 5 sources, deterministic IDs, source-url fallback, missing/null fields (including the left-join hole where `scorecardSource` is null), sort stability, slug resolution, single-file ↔ directory migration, idempotency, stale-file cleanup with un-migration, dry-run preview of removals, empty input, multi-source aggregation, valid-yaml round-trip. Adversarial input coverage for `readManagedFile` (missing/empty/non-array/malformed entries). Every branch of `formatWaveLabel`.
- [x] 188-test scorecard suite passes (`crux/lib/scorecards/__tests__` + `crux/lib/scorecard-import/__tests__`).
- [x] Full gate green (73/73 + 2 pre-existing advisories: TypeScript crux baseline, orphan-entity detection).
- [x] Full `pnpm test` green except for QUA-1062 (pre-existing groundskeeper retry test failure on main, file untouched by this PR — filed separately).
- [x] Real prod sync exercised: `pnpm crux fb sync-scorecard-facts` against prod fetched 56 grades, wrote 9 entities, skipped 16 (12 missing fb-entity yaml + 4 stableId not in entity catalog). All identifying-data shown in CLI output for operator follow-up.
- [x] `<FBF entity="anthropic" property="fli-index-grade" />` and the four other scorecard predicates render correctly via `pnpm crux fb show anthropic`.
- [x] `--no-factbase-mirror` opt-out flag verified end-to-end on `tb import-scorecards sync`.
- [x] Adversarial CLI inputs (bogus flags, shell-meta in args, special chars) handled cleanly without crash.

Closes QUA-865.


OAGr:
Pulling back from this — I think the underlying QUA-865 goal is the wrong goal, and we should close it as won't-do rather than ship the mirror.

**The framing problem.** QUA-841 flagged "grades aren't in FactBase, that's inconsistent." But FactBase is for **typed properties of the entity** (founded year, headcount, model size). A scorecard grade isn't that — it's a **third party's assessment of the entity**, with its own methodology, scale, dimensions, and version. Different kind of data, shouldn't be flattened into the same store.

Three concrete smells in this PR that point the same direction:

1. **All 5 predicates are `dataType: text`** — explicitly "to preserve heterogeneous source formats" (`C+`, `34%`, `46/100`, `Fulfilled`). We're stuffing display strings into a typed-fact store. Can't aggregate, compare, or compute on these.
2. **The "subject" isn't really the subject.** `subject=anthropic, predicate=fli-index-grade, value="C+"` reads like a property of Anthropic. It isn't — it's FLI's opinion of Anthropic in a specific report. The structure (per-dimension scores, peer rank, methodology version) gets discarded by the flattening.
3. **The grade is meaningless without context.** "C+" needs "out of what scale, by whom, when, evaluating what." `<FBF>` returns the bare letter; a bespoke component renders the context naturally.

**Proposed alternative (~30 LOC):** ship `<ScorecardGrade entity="anthropic" source="fli" />` that does a build-time fetch from wiki-server (same endpoint the `/organizations/[slug]?tab=scorecards` page already uses) and renders "FLI Index Winter 2025: C+ →" with source link, tooltip, dimension breakdown on hover. MDX authors get a better embed than `<FBF>` would have given them.

**Suggested action:** close this PR, close QUA-865 as won't-do with a pointer to a new ticket for `<ScorecardGrade>` if/when MDX embed of grades is actually requested by a page author. Right now I don't think anyone has asked for it — QUA-865 was filed by an audit, not by a page-authoring need.

OAGr:
Closing as not wanted

PR #4852, opened 2026-05-03
WIP: backfill-sources dual-judge + content cache + cron wiring

michaelr524:
## Summary (WIP — draft for diff viewing)

- Dual-judge wrapper: every quote-extraction and entailment call to Haiku/Sonnet now also runs in parallel against a local vLLM server. Haiku/Sonnet still drives every accept/reject and DB write; vLLM observations are captured for offline comparison.
- Content cache: page text passed to the judges is sha256-hashed and persisted under `dev/backfill-content-cache/<sha>.txt` so any candidate can be replayed byte-for-byte after the run.
- Report: per-record `candidates[]` now carries url/provider/decision/rejection_reason, both judges' raw text + parsed quotes + durations, and the content_sha256.
- Groundskeeper cron: new `backfill-sources-enqueue` task + handler.
- Analysis scripts under `dev/`: matched-sample audit, judge comparison, unmatched bucketing, replay-on-cached-content, uber-judge entailment tiebreaker.

## Known issue (not addressed in this PR)

Writes to `facts.source` and `policy_stakeholders.source` are wiped by the YAML→PG sync workflow (which runs after every push touching `data/entities/**` or `packages/factbase/data/fb-entities/**`). The matched URLs survive only for tables that are PG-primary (page_citations, equity_positions). Fix planned in a follow-up: write to YAML for these two tables and let sync propagate.

## Test plan

- [ ] Smoke run against prod with `--limit=2 --dry-run` produces a report containing `candidates[]` with both judges' observations
- [ ] vLLM server running on `http://localhost:8000/v1`; if unreachable, run continues with `vllm: null` fields
- [ ] Existing backfill-sources unit tests still pass (`pnpm exec vitest run crux/lib/backfill-sources`)

## Notes

This is a draft for diff review, not for merge. The branch contains dev-only commits (settings.json edits, scratch handover.md) that must be reverted before any merge.


michaelr524:
Superseded by #4870 (cleaned-up version with same feature). prod branch is strictly ahead — 0 unique commits here.

PR #4865, opened 2026-05-04
Add shared agent-end workflow

OAGr:
## Summary

- move the detailed `/agent-end` instructions into `docs/agent-workflows/agent-end.md` as the canonical cross-agent workflow
- keep `.claude/commands/agent-end.md` as a directive Claude-only adapter for built-ins like `/clear` and `/rename`
- add the Codex skill shim at `.agents/skills/source-command-agent-end`, matching the repo's existing thin-pointer convention
- document the canonical workflow location convention in `AGENTS.md`

Scoped intentionally to `agent-end`; `admin-setup` and `agent-reset` still have inline bodies and can be extracted separately.

## Test plan

- [x] `git diff --check`
- [x] static checks: Codex shim exists under `.agents/skills`, no `.claude/skills/source-command-agent-end` remains
- [x] `pnpm crux sys agent-checklist --help` to verify the documented `status`, `check --na`, and `complete` subcommands
- [x] `pnpm crux w validate gate --fix`
- [x] pre-push `pnpm crux w validate gate`
- [x] review marker flow: triage, mechanical, narrow detector, hostile/manual review, redteam/manual review, category, final

Notes: the gate passed all blocking checks. The Crux TypeScript check still reports advisory failures that appear unrelated to this documentation-only change. I did not run the full destructive `agent-end` reset flow in this live slot; the tested path is the non-destructive discovery/link/command-shape path.

## Summary by CodeRabbit

* **Documentation**
  * Added a canonical "Agent End" workflow for safely closing sessions with no PR to ship.
  * Introduced concise command/skill pointers that reference the canonical workflow rather than duplicating steps.
  * Clarified runtime-specific guidance and when to edit workflow bodies vs. runtime adapters to improve consistency.


OAGr:
A few concrete fixes. The direction (extract canonical, thin per-runtime adapters) is right; some of the execution drifts from the house convention or loses content from the original.

### 1. Move the new skill out of `.claude/skills/`

The PR creates `.claude/skills/source-command-agent-end/SKILL.md`, but per `AGENTS.md` the Codex skills directory is `.agents/skills/`. Every existing shim lives there (`.agents/skills/source-command-{batch,page-authoring,review-ontology,...}`). As filed, Codex won't find this skill and it duplicates `/agent-end` for Claude.

```
git mv .claude/skills/source-command-agent-end .agents/skills/source-command-agent-end
```

### 2. Match the existing shim body

Existing shims (e.g., `.agents/skills/source-command-batch/SKILL.md`) have a uniform 3-line body that just points at the canonical file. The new shim adds five "Important:" bullets, four of which duplicate content in `docs/agent-workflows/agent-end.md` (don't kill processes outside this slot, ask before discarding, etc.). Duplicates will drift.

Suggest replacing the shim body with the standard pattern:

```markdown
# source-command-agent-end

This skill is a thin pointer to the canonical instructions for `agent-end`.

**Read `docs/agent-workflows/agent-end.md` and follow it as written.** That file is the single source of truth for this command's behaviour and is shared across all agent runtimes (Claude Code slash command, Codex skill, etc.). Do not duplicate or paraphrase its content here.
```

If the "do not skip the AGENTS.md problem-disposition summary" reminder is genuinely missing from the canonical doc and is load-bearing, add it to the canonical doc, not to the shim.

### 3. Update `AGENTS.md` in this PR

`AGENTS.md:14` currently says: *"The `.agents/skills/source-command-*` directories are thin pointers to the canonical command files in `.claude/commands/*.md`."* This PR introduces a third location — `docs/agent-workflows/*.md` — for runtime-neutral canonical content. Without an AGENTS.md update, the next person adding a shim won't know which target is correct.

Suggested replacement for that paragraph:

> The `.agents/skills/source-command-*` directories are thin pointers to canonical command bodies. The canonical body lives in `docs/agent-workflows/<name>.md` for runtime-neutral workflows, or `.claude/commands/<name>.md` otherwise. When a workflow has been extracted to `docs/agent-workflows/`, the matching `.claude/commands/<name>.md` becomes a Claude-only adapter (it covers Claude built-ins like `/clear` and `/rename`) and defers to the canonical doc for the rest.

Also worth adding a "Canonical workflows" row or column to the runtime table so the third location is visible at a glance.

### 4. Make the Claude adapter directive, not soft

The new `.claude/commands/agent-end.md` reads "Follow the canonical workflow in [link]." That's soft — it doesn't tell the agent *what* to do with the link. Existing shims use the more directive "**Read X and follow it as written.**"

Suggested rewrite:

```markdown
---
description: End the current session — log it, update issues, clean up, reset slot
effort: low
---

# Agent End

Close out a session that does not need `/agent-ship` (research, abandoned work, maintenance, or a quick fix already pushed by hand).

**Read `docs/agent-workflows/agent-end.md` and execute every step in order.** That file is the single source of truth and is shared across agent runtimes.

Claude-specific final step: after the workflow finishes, tell the user to run `/clear` (wipe context) and `/rename` (clear the Claude Code session name). Both are Claude Code built-ins and cannot be invoked from this command file.
```

### 5. Restore the "when to use" decision table

The original had a 5-row table (`/agent-end` vs `/agent-ship`: ship a PR / research only / abandoned / quick fix already pushed / PR patrol). The new doc demotes it to one prose sentence. Tables route faster than prose, and this is exactly the kind of disposition decision an agent makes at the top of a workflow. Restore it as Section 0 of `docs/agent-workflows/agent-end.md`, before Section 1.

### 6. Tighten Step 1 prose

The original said: *"If the checklist doesn't exist (e.g., quick fix session), that's fine — skip it."* — a clear conditional. The new Step 1 says *"Validate the session checklist if it exists"* and then *"If any checklist item remains open, either complete it or mark it N/A with a reason"* — but the bash command `pnpm crux sys agent-checklist complete 2>/dev/null || true` silently swallows everything, so the agent never sees an "open item" signal. Either:

- restore the original "doesn't exist? skip" framing, or
- add an explicit `pnpm crux sys agent-checklist status` step before `complete`, with a rule like "if status shows open items, resolve them with `mark-na <id> --reason=…` or by completing them, then re-run `complete`".

(I haven't verified the exact `crux sys agent-checklist` subcommand surface — pick whatever the real one is.)

### 7. Link to the ship workflow

`docs/agent-workflows/agent-end.md` says "use the ship workflow instead" but doesn't link to it. If `docs/agent-workflows/agent-ship.md` exists or is planned, link directly. Otherwise reference `/agent-ship` (Claude) and `agent-ship` (Codex) by name.

### 8. Validate end-to-end before merging

Gate validation doesn't exercise the new indirection. Worth running once in a slot: kick off a throwaway research session, run `/agent-end`, confirm Claude opens `docs/agent-workflows/agent-end.md` and executes each step. Same for Codex if you have a Codex slot handy. A one-line "tested in slot aN, all 9 steps executed cleanly" in the PR description would close this out.

### 9. (Defer) Other shims

`admin-setup` and `agent-reset` still have inline bodies. Fine for follow-up PRs — worth a one-line note in the description that this is intentionally `agent-end`-only so reviewers don't think it was missed.

OAGr:
CodeRabbit leak-check note verified and intentionally not changed: `pnpm crux linear leak-check` is the registered command. It appears in `pnpm crux linear --help` and is wired in `crux/commands/linear.ts`. `pnpm crux sys --help` has no `linear` sys domain, so changing this to `pnpm crux sys linear leak-check` would document the wrong namespace.

